PR #4, opened 2021-04-28
Dependency Injector done

mankokolya:


freedmanMain (approved):

src/main/java/mate.academy/service/ProductService.java
```diff
@@ -1,8 +1,10 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
```
RomanDubovskyi, line 7: we use component on the interface impls instead of interfaces itself

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,76 @@
...
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> componentImplementations = new HashMap<>();
+        componentImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        componentImplementations.put(ProductParser.class, ProductParserImpl.class);
+        componentImplementations.put(ProductService.class, ProductServiceImpl.class);
+        if (interfaceClazz.isAnnotationPresent(Component.class)) {
+            return componentImplementations.get(interfaceClazz);
+        }
+        if (interfaceClazz.isInterface()) {
```
RomanDubovskyi, line 69: this check verifies that input class is an interface and if it's true we throw exception, why do we need it?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,76 @@
...
+    private Object createNewInstance(Class<?> clazz) {
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | IllegalAccessException | InstantiationException
```
RomanDubovskyi, line 55: use parent exception instead of 4 child ones

RomanDubovskyi (changes requested):

src/main/java/mate.academy/lib/Injector.java
mankokolya: I made it to check that if the interface is not marked by @Component and we want to inject some impl to it we throw an exception.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,75 @@
...
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (interfaceClazz.isAnnotationPresent(Component.class)) {
```
RomanDubovskyi, line 60: you check interface for annotation, but it should be present on the impl

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #5, opened 2021-05-03
added solution

safordog:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,76 @@
...
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private static Map<Class<?>, Class<?>> interfaceImplementations;
```
mankokolya, line 16: If I correctly understood this map shouldn't be static because it can be generated automatically in the future, we just use a simpler version.

mankokolya (approved):

safordog:
Може бути. Я також думав над цим. Дякую!

чт, 6 трав. 2021 о 13:48 Mykola Manko ***@***.***> пише:

RomanDubovskyi (approved):
Good job

PR #6, opened 2021-05-04
Di hw

Reastador:


src/main/java/mate.academy/service/FileReaderService.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 
+@Component
```
fotomassimo, line 6: If I understand correctly, in the README.md file the task requires to inject  "Interface implementations" if they are marked with @Component annotation, but not the original interfaces. So, we should put @Component annotation in the respective implementation classes.

src/main/java/mate.academy/service/ProductParser.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
```
fotomassimo, line 6: Same here

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,74 @@
...
+                }
+            }
+        }
+        if (clazzImplementationInstance == null) {
+            clazzImplementationInstance = createNewInstance(clazz);
+        }
+        return clazzImplementationInstance;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
```
fotomassimo, line 47: We should add the check if the clazz has  @component annotation, otherwise we should throw an exception (according to the requirement in the last video tutorial for this task)

fotomassimo (approved):

src/main/java/mate.academy/service/FileReaderService.java
RomanDubovskyi: > If I understand correctly, in the README.md file the task requires to inject "Interface implementations" if they are marked with @component annotation, but not the original interfaces. So, we should put @component annotation in the respective implementation classes.

that's right

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,78 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException
```
RomanDubovskyi, line 56: you can use parent exception here

RomanDubovskyi (changes requested):
fix component annotation, rest LGTM

RomanDubovskyi (changes requested):
you didn't fix Component annotation, it should on the implementations, not interfaces 
image

PR #7, opened 2021-05-04
first solution

Vlad-Andr:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,80 @@
...
+                        field.setAccessible(true);
+                        field.set(classImplementationInstance, fieldInstance);
+                    } catch (IllegalAccessException e) {
+                        throw new RuntimeException("Can't initialize field " + field.getName());
+                    }
+                }
+            }
+        } else {
+            throw new WrongInitializingException("Class " + clazz.getName()
+                    + " without annotation");
```
divertitoS, line 44: I think that you can simplify somehow this if-else construction with exception inside else-block

divertitoS (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,80 @@
...
+        }
+        if (classImplementationInstance == null) {
+            classImplementationInstance = createNewInstance(clazz);
+        }
+        return classImplementationInstance;
```
Max-Pochepets, line 49:
```suggestion
            }
            if (classImplementationInstance == null) {
                classImplementationInstance = createNewInstance(clazz);
            }
            return classImplementationInstance;
        }
        throw new WrongInitializingException("Class " + clazz.getName()
                + " without annotation");
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,80 @@
...
+
+    private Class<?> findImplementation(Class<?> interfaceInstance) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
+        if (interfaceInstance.isInterface()) {
+            return interfaceImplementations.get(interfaceInstance);
+        } else {
+            return interfaceInstance;
```
Max-Pochepets, line 77:
```suggestion
        if (interfaceInstance.isInterface()) {
            return interfaceImplementations.get(interfaceInstance);
        }
        return interfaceInstance;
```

Max-Pochepets (approved):

PR #8, opened 2021-05-04
initial commit

fotomassimo:


Reastador (approved):

src/main/java/mate.academy/service/impl/ProductServiceImpl.java
```diff
@@ -2,13 +2,17 @@
...
 import mate.academy.model.Product;
 import mate.academy.service.FileReaderService;
 import mate.academy.service.ProductParser;
 import mate.academy.service.ProductService;
 
+@Component
 public class ProductServiceImpl implements ProductService {
+    @Component
     private ProductParser productParser;
+    @Component
```
RomanDubovskyi, line 15:
```suggestion
    @Inject
```
Inject is used above the field you want to be initialized

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,73 @@
...
+                            + interfaceImplClass.getName(), e);
+                }
+            }
+        }
+        return interfaceClassInstance == null ? createNewInstance(interfaceImplClass)
+                : interfaceClassInstance;
+    }
+
+    private Object createNewInstance(Class<?> interfaceClassImpl) {
+        if (!interfaceClassImpl.isAnnotationPresent(Component.class)) {
```
RomanDubovskyi, line 45: let's move this check to findImplementation method.

RomanDubovskyi (changes requested):

src/main/java/mate.academy/lib/Injector.java
fotomassimo: done

src/main/java/mate.academy/service/impl/ProductServiceImpl.java
fotomassimo: done

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,73 @@
...
+    private Object createNewInstance(Class<?> interfaceClassImpl) {
+        if (instances.containsKey(interfaceClassImpl)) {
+            return instances.get(interfaceClassImpl);
+        }
+        try {
+            Object instance = interfaceClassImpl.getConstructor().newInstance();
+            instances.put(interfaceClassImpl, instance);
+            return instance;
+        } catch (InstantiationException | NoSuchMethodException
+                | InvocationTargetException | IllegalAccessException e) {
```
KseniiaMakarova, line 53:
```suggestion
        } catch (ReflectiveOperationException e) {
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,73 @@
...
+
+    private Class<?> findImplementation(Class<?> interfaceClass) {
+        Map<Class<?>, Class<?>> interfaceImplementation = new HashMap<>();
+        interfaceImplementation.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementation.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementation.put(ProductService.class, ProductServiceImpl.class);
+        Class<?> interfaceClassImpl = interfaceClass.isInterface()
+                ? interfaceImplementation.get(interfaceClass)
+                : interfaceClass;
+        if (!interfaceClassImpl.isAnnotationPresent(Component.class)) {
```
KseniiaMakarova, line 67: what if `interfaceClass` is not an in interface? we won't need to put a `@Component` annotation over it then, but this code will throw an exception

KseniiaMakarova (changes requested):

src/main/java/mate.academy/lib/Injector.java
fotomassimo: done

PR #10, opened 2021-05-05
Injector injected (lul)

LayRipper:


DeStinyShell (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,74 @@
...
+        } catch (InstantiationException | IllegalAccessException
+                | InvocationTargetException | NoSuchMethodException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> implementationMap = new HashMap<>();
+        implementationMap.put(FileReaderService.class, FileReaderServiceImpl.class);
+        implementationMap.put(ProductParser.class, ProductParserImpl.class);
```
RomanDubovskyi, line 68: what about ProductService?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,74 @@
...
+        if (interfaceClazz.isInterface()) {
```
RomanDubovskyi, line 69: not sure if we need this if here. Theoritaclly we could use this not only for interfaces.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,74 @@
...
+
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (InstantiationException | IllegalAccessException
```
RomanDubovskyi, line 59: replace these for e with parent exception that they inherit

RomanDubovskyi (changes requested):

RomanDubovskyi (changes requested):
.

RomanDubovskyi (approved):

PR #12, opened 2021-05-05
first solution

dvdovynets:


freedmanMain (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,74 @@
...
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create new instance of " + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementation = new HashMap<>();
+        interfaceImplementation.put(ProductService.class, ProductServiceImpl.class);
```
Max-Pochepets, line 65: if you hardcoded implementations map, let's check if needed one will have annotation `Component`

Max-Pochepets (changes requested):
good job

src/main/java/mate.academy/lib/Injector.java
dvdovynets: Done

Max-Pochepets (approved):
👍

PR #18, opened 2021-05-06
Solution Injector

vuyko23:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,77 @@
...
+        if (createdObjects.containsKey(clazz)) {
+            return createdObjects.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            createdObjects.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InstantiationException
+                | IllegalAccessException | InvocationTargetException e) {
```
Dalai-S, line 56: You can use "ReflectiveOperationException" here

src/main/java/mate.academy/service/impl/FileReaderServiceImpl.java
```diff
@@ -4,8 +4,10 @@
 import java.io.IOException;
 import java.nio.file.Files;
 import java.util.List;
+import mate.academy.lib.Component;
 import mate.academy.service.FileReaderService;
 
+@Component
```
Dalai-S, line 10: I think every interface should be marked(but not a class), because if you will make new class and implement this interface you will need to write this annotation again. So as for me better to write Component for interface

Dalai-S (approved):

src/main/java/mate.academy/service/impl/FileReaderServiceImpl.java
RomanDubovskyi: it's opposite, we mark impls of interface with this annotation to mark what instances should be crated

src/main/java/mate.academy/lib/Injector.java
RomanDubovskyi: \+

RomanDubovskyi (approved):

PR #19, opened 2021-05-06
Finish hw requirements.

dant0n:


VladislavTetar (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,79 @@
...
     public static Injector getInjector() {
         return injector;
     }
+
+    public Object getInstance(Class<?> interfaceClazz) {
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        if (clazz.getInterfaces().length > 0
+                && !clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Can't create instance of a " + clazz.getSimpleName()
```
Max-Pochepets, line 27: this message is informative. if you're willing to keep it like this, you should check for `Component` annotation inside `findImplementation()`. Personally I would thought that we can't create instance because there is no annotation, but it's also possible when there no implementations of given interface

Max-Pochepets (approved):

PR #21, opened 2021-05-06
patch 1.0.0

IhorRome:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,83 @@
...
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instanceMap.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException
+                | InstantiationException
+                | IllegalAccessException
+                | InvocationTargetException e) {
+            throw new RuntimeException("can't create instance", e);
```
DanyloIT, line 80: Better to add more info (for example class name)

DanyloIT (approved):
all good, only informativity of exceptions

RomanDubovskyi (approved):

PR #22, opened 2021-05-07
created Injector

Iryna-U:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,79 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz,instance);
+            return instance;
+        } catch (NoSuchMethodException | IllegalAccessException
+                | InstantiationException | InvocationTargetException e) {
```
Dalai-S, line 59: You can use "ReflectiveOperationException" here

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,79 @@
...
+        interfaceImplementation.put(FileReaderService.class, FileReaderServiceImpl.class);
+
+        if (interfaceImplementation.get(interfaceClazz).isAnnotationPresent(Component.class)) {
+            return interfaceImplementation.get(interfaceClazz);
+        }
+        if (interfaceClazz.isInterface()) {
+            throw new RuntimeException(interfaceClazz.getName() + " interface doesn't have: "
+                    + Component.class + " annotation");
+        }
+        return interfaceClazz;
```
Dalai-S, line 77: Do you really need all this code? I think you can throw exception only 
```suggestion
        throw new RuntimeException(interfaceClazz.getName() + " interface doesn't have: "
                 + Component.class + " annotation");
        
```

Dalai-S (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,79 @@
...
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementation = new HashMap<>();
+        interfaceImplementation.put(ProductService.class, ProductServiceImpl.class);
+        interfaceImplementation.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementation.put(FileReaderService.class, FileReaderServiceImpl.class);
+
+        if (interfaceImplementation.get(interfaceClazz).isAnnotationPresent(Component.class)) {
+            return interfaceImplementation.get(interfaceClazz);
+        }
+        if (interfaceClazz.isInterface()) {
```
RomanDubovskyi, line 73: this if checks if input class is an interface, why we throw exception that annotation is missing whe check failes?

src/main/java/mate.academy/lib/Injector.java
RomanDubovskyi: \+

RomanDubovskyi (changes requested):
fix check in the findImplementation method, rest LGTM

src/main/java/mate.academy/lib/Injector.java
Iryna-U: I agree. I changed the method and was inattentive.

PR #26, opened 2021-05-07
First solution

Ihor-Nykyforov:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,68 @@
...
+    private Object createNeInstance(Class<?> clazz) {
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Object instance = clazz.getConstructor().newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InstantiationException
+                | InvocationTargetException | IllegalAccessException e) {
```
Den4eg007, line 53: Use one exception that all of these inherit

Den4eg007 (commented):
GJ

src/main/java/mate.academy/lib/Injector.java
Den4eg007, line 53: Use one exception that all of these inherit

Den4eg007 (changes requested):

Den4eg007 (approved):

RomanDubovskyi (approved):

PR #37, opened 2021-05-08
added solution

PepsiTheDoge:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,77 @@
...
+        Class<?> clazz = findImplementation(interfaceClazz);
+        Field[] declaredFields = clazz.getDeclaredFields();
+        Object clazzImplementationInstance = null;
+        for (Field field : declaredFields) {
+            if (field.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = getInstance(field.getType());
+                clazzImplementationInstance = createNewInstance(clazz);
+                field.setAccessible(true);
+                try {
+                    field.set(clazzImplementationInstance, fieldInstance);
```
KseniiaMakarova, line 33:
```suggestion
                try {
                    field.setAccessible(true);
                    field.set(clazzImplementationInstance, fieldInstance);
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,77 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InstantiationException
+                | IllegalAccessException | InvocationTargetException e) {
```
KseniiaMakarova, line 72:
```suggestion
        } catch (ReflectiveOperationException e) {
```

KseniiaMakarova (approved):

PR #39, opened 2021-05-08
1 added solution

SerhiiKysliakov:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,71 @@
...
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
+        if (interfaceClazz.isInterface()) {
+            return interfaceImplementations.get(interfaceClazz);
+        }
+        return interfaceClazz;
```
Anatolii-Zarytskyi, line 69: here you need to check class with Component Annotation

Anatolii-Zarytskyi (changes requested):

src/main/java/mate.academy/Main.java
```diff
@@ -4,13 +4,15 @@
...
+import mate.academy.service.impl.ProductServiceImpl;
 
 public class Main {
 
     public static void main(String[] args) {
         // Please test your Injector here. Feel free to push this class as a part of your solution
         Injector injector = Injector.getInjector();
-        ProductService productService = null;
+        ProductService productService = (ProductService) injector
+                .getInstance(ProductServiceImpl.class);
```
Resci, line 15: We should pass to the injector abstraction or implementation?

Resci (changes requested):

PR #40, opened 2021-05-08
done

ritakisluhina:


src/main/java/mate.academy/service/FileReaderService.java
```diff
@@ -3,5 +3,6 @@
 import java.util.List;
 
 public interface FileReaderService {
+    @Component
```
Aalderaan, line 6: There is no need to mark the method. With @Component we need to mark the class which implements interface.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,73 @@
...
+    private Class<?> findImplementation(Class<?> interfaceClass) {
+        Map<Class<?>, Class<?>> interfaceImpl = new HashMap<>();
+        interfaceImpl.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImpl.put(ProductService.class, ProductServiceImpl.class);
+        interfaceImpl.put(ProductParser.class, ProductParserImpl.class);
+        if (interfaceClass.isInterface()) {
+            return interfaceImpl.get(interfaceClass);
+        }
+        return interfaceClass;
+    }
```
Aalderaan, line 72: Maybe it is better to add 'if' statement, that checks if the class is marked with @Component annotation. Otherwise:  RuntimeExeption, telling that "This class is not marked with 'Component' annotation" ?

Aalderaan (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,73 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InstantiationException
+                | IllegalAccessException | InvocationTargetException e) {
```
Den4eg007, line 58: You can use one exception that all of these inherit

Den4eg007 (changes requested):

src/main/java/mate.academy/service/FileReaderService.java
ritakisluhina: done

Aalderaan (approved):

src/main/java/mate.academy/Main.java
```diff
@@ -4,13 +4,15 @@
...
+import mate.academy.service.impl.ProductServiceImpl;
 
 public class Main {
 
     public static void main(String[] args) {
         // Please test your Injector here. Feel free to push this class as a part of your solution
         Injector injector = Injector.getInjector();
-        ProductService productService = null;
+        ProductService productService =
+                (ProductService) injector.getInstance(ProductServiceImpl.class);
```
Resci, line 15: We should pass to the injector abstraction or implementation?

Resci (changes requested):

PR #43, opened 2021-05-10
added Injector

Alina-Vasylyshyn:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,80 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InstantiationException
+                | InvocationTargetException | IllegalAccessException e) {
```
dvdovynets, line 57: You need change those four exceptions to ReflectiveOperationException because they all extends from it.

dvdovynets (changes requested):

src/main/java/mate.academy/lib/Injector.java
Alina-Vasylyshyn: Done

Resci (approved):

PR #44, opened 2021-05-10
Ready to review.

MakeDreamCode:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,77 @@
...
+            }
+        }
+        if (clazzImplementationInstance == null) {
+            clazzImplementationInstance = createNewInstance(clazz);
+        }
+        return clazzImplementationInstance;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
+        if (!clazz.isAnnotationPresent(Component.class)) {
```
mankokolya, line 48: You have to be able to initialize an instance of a regular class that not implement any interface. In your case, you discard all classes that are not marked by the @Component. Try to add an extra check  if clazz is interface && !clazz.isAnnPresent

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,9 +1,77 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | IllegalAccessException
+                | InstantiationException | InvocationTargetException e) {
```
mankokolya, line 62: You are not using each exception. You can change all of them to one common Exception from which they are inherited

mankokolya (changes requested):

src/main/java/mate.academy/lib/Injector.java
MakeDreamCode: Replaced by ReflectiveOperationException

mankokolya (approved):

PR #48, opened 2021-05-10
created injector and annotations

Roman0607:


src/main/java/mate.academy/service/ProductParser.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
```
Anatolii-Zarytskyi, line 6: Do we need annotation here?

src/main/java/mate.academy/service/impl/ProductParserImpl.java
```diff
@@ -5,7 +5,6 @@
 import mate.academy.service.ProductParser;
 
 public class ProductParserImpl implements ProductParser {
```
Anatolii-Zarytskyi, line 7: Do we need annotation here? think about it

src/main/java/mate.academy/service/impl/ProductServiceImpl.java
```diff
@@ -2,13 +2,16 @@
 
 import java.util.List;
 import java.util.stream.Collectors;
+import mate.academy.lib.Inject;
 import mate.academy.model.Product;
 import mate.academy.service.FileReaderService;
 import mate.academy.service.ProductParser;
 import mate.academy.service.ProductService;
 
 public class ProductServiceImpl implements ProductService {
```
Anatolii-Zarytskyi, line 11: same

Anatolii-Zarytskyi (changes requested):

src/main/java/mate.academy/Main.java
```diff
@@ -10,7 +10,8 @@
     public static void main(String[] args) {
         // Please test your Injector here. Feel free to push this class as a part of your solution
         Injector injector = Injector.getInjector();
-        ProductService productService = null;
+        ProductService productService = (ProductService)
+                injector.getInstance(ProductService.class);
```
Resci, line 14:
```suggestion
        ProductService productService = (ProductService) injector.getInstance(ProductService.class);
```

src/main/java/mate.academy/service/impl/ProductParserImpl.java
Resci: What about FileReaderService?

Resci (changes requested):

Resci (approved):

PR #54, opened 2021-07-02
HW: DI

SakVladyslav:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,80 @@
...
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
+
+        if (interfaceImplementations.get(interfaceClazz).isAnnotationPresent(Component.class)) {
+            return interfaceImplementations.get(interfaceClazz);
+        }
+
+        if (!interfaceClazz.isInterface()) {
+            return interfaceClazz;
+        }
```
KseniiaMakarova, line 74: let's maybe change the order of these 2 `if`s?
if it won't be an interface, we will get NPE on line 68

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,80 @@
...
+            return instances.get(clazz);
+        }
+
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (InstantiationException | IllegalAccessException
+                | InvocationTargetException | NoSuchMethodException e) {
```
KseniiaMakarova, line 57:
```suggestion
 } catch (ReflectiveOperationException e) {
```

KseniiaMakarova (changes requested):
great job! just fix 1 small thing

KseniiaMakarova (approved):

PR #59, opened 2021-07-07
Injector implemented. Tests passed.

nikdagaev:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -8,6 +20,57 @@ public static Injector getInjector() {
...
+        Constructor<?> constructor;
+        Object instance;
+        try {
+            constructor = clazz.getConstructor();
+            instance = constructor.newInstance();
+        } catch (InstantiationException
+                | NoSuchMethodException
+                | InvocationTargetException
+                | IllegalAccessException e) {
+            throw new RuntimeException("Can't create object of this class: " + clazz);
```
Veronika-Alekseienko, line 57:
```suggestion
            throw new RuntimeException("Can't create object of this class: " + clazz.getName());
```

src/main/java/mate.academy/Main.java
```diff
@@ -4,13 +4,15 @@
...
+import mate.academy.service.impl.ProductServiceImpl;
 
 public class Main {
 
     public static void main(String[] args) {
         // Please test your Injector here. Feel free to push this class as a part of your solution
         Injector injector = Injector.getInjector();
-        ProductService productService = null;
+        ProductService productService =
+                (ProductServiceImpl) injector.getInstance(ProductServiceImpl.class);
```
Veronika-Alekseienko, line 15: i think better use abstraction , not impl 
```suggestion
                (ProductService) injector.getInstance(ProductService.class);
```

Veronika-Alekseienko (changes requested):

Veronika-Alekseienko (approved):
Well done!)

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -8,6 +20,56 @@ public static Injector getInjector() {
...
+    private Object createNewInstance(Class<?> clazz) {
+        Constructor<?> constructor;
+        Object instance;
+        try {
+            constructor = clazz.getConstructor();
+            instance = constructor.newInstance();
+        } catch (InstantiationException
+                | NoSuchMethodException
+                | InvocationTargetException
+                | IllegalAccessException e) {
```
KseniiaMakarova, line 56:
```suggestion
        } catch (ReflectiveOperationException e) {
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -8,6 +20,56 @@ public static Injector getInjector() {
...
+            classImplementationInstance = createNewInstance(clazz);
+        }
+        return classImplementationInstance;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
+        Constructor<?> constructor;
+        Object instance;
+        try {
+            constructor = clazz.getConstructor();
```
KseniiaMakarova, line 51: should we first check the `instances` map? what if we already have created such an object?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -8,6 +20,56 @@ public static Injector getInjector() {
...
+        return instance;
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
+        if (interfaceClazz.isInterface()) {
+            return interfaceImplementations.get(interfaceClazz);
```
KseniiaMakarova, line 68: you just return an implementation from the map here, but you should also check if it has the `@Component` annotation

KseniiaMakarova (changes requested):

src/main/java/mate.academy/lib/Injector.java
nikdagaev: I didn't create this map in my implementation because it's a simple parser and I think the logic will be overcompicated in this case if we will use instances map

KseniiaMakarova (changes requested):
let's still create the `instances map`
the logic won't be overcomplicated, but we will not run the same logic multiple times if we need several objects of some type

KseniiaMakarova (approved):

PR #60, opened 2021-07-07
wrote own injector

Veronika-Alekseienko:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+            for (Field field : declaredFields) {
+                if (field.isAnnotationPresent(Inject.class)) {
+                    Object fieldInstance = getInstance(field.getType());
+                    clazzImplementationInstance = createNewInstance(clazz);
+                    field.setAccessible(true);
+                    try {
+                        field.set(clazzImplementationInstance, fieldInstance);
+                    } catch (IllegalAccessException e) {
+                        throw new RuntimeException("Can't initialize value. "
+                                + "Class: " + clazz.getName()
```
nikdagaev, line 37: "Can't initialize value. " + "Class: " can be in one string

nikdagaev (changes requested):

nikdagaev (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InstantiationException
+                | IllegalAccessException | InvocationTargetException e) {
```
KseniiaMakarova, line 60:
```suggestion
        } catch (ReflectiveOperationException e) {
```
we may use this for simplicity

KseniiaMakarova (approved):

PR #61, opened 2021-07-07
hw-20 added solution

OlesiaSmitiukh:


PashaTerletskyi (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        Field[] declaredFields = clazz.getDeclaredFields();
+        for (Field field : declaredFields) {
+            if (field.isAnnotationPresent(Component.class)) {
```
KseniiaMakarova, line 28: why Component?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Class: " + clazz.getName() + " is unsupported");
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InstantiationException | IllegalAccessException
+                | InvocationTargetException e) {
```
KseniiaMakarova, line 59:
```suggestion
        } catch (ReflectiveOperationException e) {
```

KseniiaMakarova (changes requested):

src/main/java/mate.academy/lib/Injector.java
OlesiaSmitiukh: ok

src/main/java/mate.academy/lib/Injector.java
OlesiaSmitiukh: here should be Inject

KseniiaMakarova (approved):

PR #62, opened 2021-07-07
implemented Injector, added annotations Component and Inject

Daniel1004K:


Tamara-Paskalova (approved):

src/main/java/mate.academy/service/ProductService.java
```diff
@@ -1,8 +1,10 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
```
KseniiaMakarova, line 7: you should place this annotation over the implementations, and not over the interfaces

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,72 @@
...
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        if (!interfaceClazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Its not possible to create instance of "
+                    + interfaceClazz.getName());
+        }
```
KseniiaMakarova, line 33: we should check the implementation, and not the interfaceClazz

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,72 @@
...
+    private Object createNewInstance(Class<?> clazz) {
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Object instance = clazz.getConstructor().newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InstantiationException
+                | IllegalAccessException | InvocationTargetException e) {
```
KseniiaMakarova, line 61:
```suggestion
        } catch (ReflectiveOperationException e) {
```

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #64, opened 2021-07-08
Added injector implementation and Two new annotations: @Component - i…

Optipist:
…t says about component class, @Inject - it says about field that must be injected

### DIFF
--- a/src/main/java/mate.academy/Main.java
+++ b/src/main/java/mate.academy/Main.java
@@ -4,13 +4,15 @@ import java.util.List;
 import mate.academy.lib.Injector;
 import mate.academy.model.Product;
 import mate.academy.service.ProductService;
+import mate.academy.service.impl.ProductServiceImpl;
 
 public class Main {
 
     public static void main(String[] args) {
-        // Please test your Injector here. Feel free to push this class as a part of your solution
         Injector injector = Injector.getInjector();
-        ProductService productService = null;
+        ProductService productService
+                = (ProductService) injector.getInstance(
+                        ProductServiceImpl.class);
         List<Product> products = productService.getAllFromFile("products.txt");
         products.forEach(System.out::println);
     }
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,13 +1,58 @@
 package mate.academy.lib;
 
+import java.lang.reflect.Field;
+import mate.academy.storage.ImplementationStorageImpl;
+import mate.academy.storage.InstanceStorageImpl;
+import mate.academy.storage.Storage;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private static Storage<Class<?>, Class<?>> implementations
+            = new ImplementationStorageImpl();
+    private static Storage<Class<?>, Object> instances
+            = new InstanceStorageImpl();
 
     public static Injector getInjector() {
         return injector;
     }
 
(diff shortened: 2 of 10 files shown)


aaananas (approved):

src/main/java/mate.academy/Main.java
```diff
@@ -4,13 +4,15 @@
...
 
 public class Main {
 
     public static void main(String[] args) {
-        // Please test your Injector here. Feel free to push this class as a part of your solution
         Injector injector = Injector.getInjector();
-        ProductService productService = null;
+        ProductService productService
+                = (ProductService) injector.getInstance(
+                        ProductServiceImpl.class);
```
KseniiaMakarova, line 15:
```suggestion
        ProductService productService
                = (ProductService) injector.getInstance(ProductServiceImpl.class);
```

KseniiaMakarova, line 15: you should pass the interface class here - not the implementation class

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,58 @@
...
+import mate.academy.storage.ImplementationStorageImpl;
+import mate.academy.storage.InstanceStorageImpl;
+import mate.academy.storage.Storage;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private static Storage<Class<?>, Class<?>> implementations
+            = new ImplementationStorageImpl();
+    private static Storage<Class<?>, Object> instances
+            = new InstanceStorageImpl();
```
KseniiaMakarova, line 13: for future: there is no actual need to wrap java collections into some oop form
we can use them in the original form (map, list, so on)

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,58 @@
...
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        Field[] fields = interfaceClazz.getDeclaredFields();
```
KseniiaMakarova, line 22: should we look through interface fields or class fields?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,58 @@
...
+        throw new RuntimeException("This class are without @Component annotation\n"
+                + clazz.getName());
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
+        return instances.get(clazz);
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (((ImplementationStorageImpl)implementations)
```
KseniiaMakarova, line 52: here you are using exact implementation
let's change the Storage interface so that we can remove casting here

KseniiaMakarova (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -50,7 +50,7 @@ private Object createNewInstance(Class<?> clazz) {
 
     private Class<?> findImplementation(Class<?> interfaceClazz) {
         if (((ImplementationStorageImpl)implementations)
```
KseniiaMakarova, line 52: we are still casting - let's remove casting

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #68, opened 2021-07-08
added solution

Mzuha:


Basilius-Third (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (InvocationTargetException
+                | InstantiationException
+                | IllegalAccessException
+                | NoSuchMethodException e) {
```
KseniiaMakarova, line 62:
```suggestion
        } catch (ReflectiveOperationException e) {
```

KseniiaMakarova (approved):

PR #71, opened 2021-07-08
HW-Injector

docer95:
create Injector
add notification @Inject and @Component in classes Impl
fix Main

### DIFF
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,13 +1,74 @@
 package mate.academy.lib;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.util.HashMap;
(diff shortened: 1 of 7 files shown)


Basilius-Third (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                | InstantiationException | IllegalAccessException e) {
```
KseniiaMakarova, line 59:
```suggestion
        } catch (ReflectiveOperationException e) {
```

KseniiaMakarova (approved):

PR #79, opened 2021-09-07
Hw dependency injection

AlexRozv:


kovalskiy7777 (approved):
well done!!!

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        Field[] declaredFields = clazz.getDeclaredFields();
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Can't create this component: " + clazz.getName());
+        }
```
MaliukDaria, line 29: It seems to me more logical to carry out this check in the findImplementation method. If there is no Component annotation above class, than we wont return it's implementation

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                | InstantiationException | IllegalAccessException e) {
+            throw new RuntimeException("Can't create new instance of " + clazz.getName());
+        }
+
```
MaliukDaria, line 62: redundant empty line

MaliukDaria (changes requested):

src/main/java/mate.academy/lib/Injector.java
AlexRozv: fixed

src/main/java/mate.academy/lib/Injector.java
AlexRozv: fixed

MaliukDaria (approved):
🔥

MaliukDaria (approved):

PR #81, opened 2021-09-11
Injection homework

mishamolnar:


src/main/java/mate.academy/service/impl/ProductParserImpl.java
```diff
@@ -21,6 +22,7 @@ public Product parse(String productInfo) {
         product.setCategory(data[CATEGORY_POSITION]);
         product.setDescription(data[DESCRIPTION_POSITION]);
         product.setPrice(new BigDecimal(data[PRICE_POSITION]));
+
```
Sinelife, line 25: lets delete empty line

Sinelife (changes requested):

Sinelife (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        Field[] declaredFields = clazz.getDeclaredFields();
+        for (Field field : declaredFields) {
+            if (!clazz.isAnnotationPresent(Component.class)) {
+                throw new RuntimeException("Can't create an instance of the class"
+                        + clazz.getName());
+            }
```
MaliukDaria, line 31: we can move this check to findImplementation method. If there is no Component annotation above the class, then we shouldn't return its implementation

MaliukDaria (changes requested):

MaliukDaria (approved):
🚀

MaliukDaria (approved):

PR #86, opened 2021-09-19
jv-dependency-injection implemented

lavrivskyi:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,70 @@
...
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private Map<Class<?>, Object> instances = new HashMap<>();
```
kovalskiy7777, line 17:
```suggestion
    private final Map<Class<?>, Object> instances = new HashMap<>();
```

kovalskiy7777 (approved):

src/main/java/mate.academy/lib/Injector.java
lavrivskyi: Fixed.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,70 @@
...
+                Constructor<?> constructor = clazz.getConstructor();
+                Object instance = constructor.newInstance();
+                instances.put(clazz, instance);
+                return instance;
+            } catch (NoSuchMethodException | InvocationTargetException
+                    | InstantiationException | IllegalAccessException e) {
+                throw new RuntimeException("Can't create new instance of " + clazz.getName());
+            }
+        }
+        throw new RuntimeException("Can't create new instance of " + clazz.getName());
```
MaliukDaria, line 60:
```suggestion
        throw new RuntimeException("Can't create new instance of " + clazz.getName());
```

lets add some information why we cant create new instance(because annotation `Component` is absent)

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,70 @@
...
+            return instances.get(clazz);
+        }
+        if (clazz.isAnnotationPresent(Component.class)) {
+            try {
+                Constructor<?> constructor = clazz.getConstructor();
+                Object instance = constructor.newInstance();
+                instances.put(clazz, instance);
+                return instance;
+            } catch (NoSuchMethodException | InvocationTargetException
+                    | InstantiationException | IllegalAccessException e) {
```
MaliukDaria, line 56: Can we replace this 4 exception with common parent?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,70 @@
...
+        }
+        throw new RuntimeException("Can't create new instance of " + clazz.getName());
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        return interfaceImplementations.get(interfaceClazz);
```
MaliukDaria, line 68: what if out input parameter `interfaceClazz` it's not an interface but a class?

MaliukDaria (changes requested):

src/main/java/mate.academy/lib/Injector.java
lavrivskyi: Fixed.

src/main/java/mate.academy/lib/Injector.java
lavrivskyi: Fixed.

src/main/java/mate.academy/lib/Injector.java
lavrivskyi: Fixed.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+        }
+        throw new RuntimeException("Can't create new instance of " + clazz.getName()
+                + "annotation \"Component\" is absent");
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (!interfaceClazz.isInterface()) {
+            throw new RuntimeException(interfaceClazz.getName()
+                    + "is not an interface! Please provide valid interface.");
+        }
```
MaliukDaria, line 66: We dont need to throw exception in case is interfaceClazz is class, we can just return this class
```suggestion
        if (!interfaceClazz.isInterface()) {
            return interfaceClazz;
        }
```

MaliukDaria (changes requested):

src/main/java/mate.academy/lib/Injector.java
lavrivskyi: Fixed.

MaliukDaria (approved):

MaliukDaria (approved):

PR #89, opened 2021-09-22
Implemented Injector

Radikhovskiy:


src/main/java/mate.academy/Main.java
```diff
@@ -4,13 +4,13 @@
...
 
 public class Main {
-
     public static void main(String[] args) {
-        // Please test your Injector here. Feel free to push this class as a part of your solution
         Injector injector = Injector.getInjector();
-        ProductService productService = null;
+        ProductService productService = (ProductService) injector
+                .getInstance(ProductServiceImpl.class);
         List<Product> products = productService.getAllFromFile("products.txt");
```
Sinelife, line 14: throws NullPointerException

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,68 @@
...
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementationInstance(interfaceClazz);
+        Field[] declaredFields = clazz.getDeclaredFields();
+        for (Field field: declaredFields) {
+            if (field.isAnnotationPresent(Inject.class)) {
+                clazzImplementationInstance = createNewInstance(clazz);
+            }
+        }
```
Sinelife, line 31: here you create instance of clazz but you didnt set value to field.
Because of that fields won't be initialized and you will get NullPointerException.
Rewatch this part of Bogdan's video https://youtu.be/7_h6nD529c0?t=619

Sinelife (changes requested):

Sinelife (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,81 @@
...
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        if (!(interfaceClazz.isInterface())
+                && !(interfaceClazz.isAnnotationPresent(Component.class))) {
+            throw new RuntimeException("Class is unsupported.");
```
Max-Pochepets, line 26: If I'll get a message like this, I would not know how to fix this.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,81 @@
...
+                            + clazz.getName());
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                | InstantiationException | IllegalAccessException e) {
+            throw new RuntimeException("Can't create new instance of " + clazz.getName());
```
Max-Pochepets, line 67: there is the same behaviour for all these exceptions. they can be composed into one

Max-Pochepets (changes requested):

Max-Pochepets (approved):

Max-Pochepets (changes requested):
test

Max-Pochepets (approved):

PR #90, opened 2021-09-22
injector-hm solution

kryzhen:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -8,6 +20,47 @@ public static Injector getInjector() {
...
+        if (interfaceClazz.isAnnotationPresent(Component.class)) {
+            clazzImplInstance = createNewInstance(clazz);
+            for (Field field : declaredFields) {
+                if (field.isAnnotationPresent(Inject.class)) {
+                    clazzImplInstance = createNewInstance(clazz);
+                    Object fieldInstance = getInstance(field.getType());
+                    try {
+                        field.set(clazzImplInstance, fieldInstance);
+                    } catch (IllegalAccessException e) {
+                        e.printStackTrace();
```
RomanDubovskyi, line 35: e.printStackTrace is a bad practice, throw an exception instead

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -8,6 +20,47 @@ public static Injector getInjector() {
...
+            return interfaceImpl.get(interfaceClazz);
+        }
+        return interfaceClazz;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            return constructor.newInstance();
+        } catch (InstantiationException | InvocationTargetException
```
RomanDubovskyi, line 61: we can replace these 4 exc with their common parent

src/main/java/mate.academy/service/ProductService.java
```diff
@@ -1,8 +1,10 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
```
RomanDubovskyi, line 7: This annotation should be present only on implementation classes to help us find a class for object creation

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #95, opened 2021-09-23
Injection class was created

Koriastras:


src/main/java/mate.academy/service/impl/ProductParserImpl.java
```diff
@@ -1,9 +1,11 @@
 package mate.academy.service.impl;
 
 import java.math.BigDecimal;
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 import mate.academy.service.ProductParser;
 
+@Component
 public class ProductParserImpl implements ProductParser {
```
Yehor-Zapara, line 10: Let's delete whitespace

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,82 @@
...
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object classImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        checkAnnotation(clazz);
+        Field[] fieldsDeclaration = clazz.getDeclaredFields();
```
Yehor-Zapara, line 27: Maybe better to rename variable **fieldsDeclaration** to **declaredFields** or **fields**. **fieldsDeclaration** is confusing because we don't declare fields in this row, we get it and try to initialize them below, they are already declared in classes

Yehor-Zapara (changes requested):

Yehor-Zapara (approved):

Max-Pochepets (approved):

Max-Pochepets (changes requested):
test

Max-Pochepets (approved):

PR #96, opened 2021-09-23
task injection; were inmplimentated @Component and @Inject and class …

Yelinskyi:
…Injector

### DIFF
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,13 +1,77 @@
 package mate.academy.lib;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.util.HashMap;
+import java.util.Map;
+import mate.academy.service.FileReaderService;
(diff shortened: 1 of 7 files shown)


VladyslavLekhter (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                | InstantiationException | IllegalAccessException e) {
```
Max-Pochepets, line 62: There is the same behaviour for all these exceptions. They can be composed into one exception.

Max-Pochepets (changes requested):

Max-Pochepets (approved):

Max-Pochepets (changes requested):
test

Max-Pochepets (approved):

PR #97, opened 2021-09-24
Java_DI_HW

AlexPravda1:
- @Inject annotation is checked
- @Component annotation is checked
- injector pushed in main method
- interface and Impl classes MAP is created manually


RomanDubovskyi (approved):
GJ!

PR #99, opened 2021-09-24
realized jv-dependency-injection App

sashahordenin:


Sinelife (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                | InstantiationException | IllegalAccessException e) {
```
Max-Pochepets, line 61: There is the same behaviour for all these exceptions. They can be composed into one.

Max-Pochepets (changes requested):

src/main/java/mate.academy/lib/Injector.java
sashahordenin: Changed for ReflectiveOperationException

Max-Pochepets (approved):

Max-Pochepets (changes requested):
test

Max-Pochepets (approved):

PR #100, opened 2021-09-24
ST dependency injection

Evgen-Krasnenkov:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,82 @@
...
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        if (!interfaceClazz.equals(ProductServiceImpl.class)
+                && !interfaceClazz.equals(ProductService.class)
+                && !interfaceClazz.equals(ProductParser.class)
+                && !interfaceClazz.equals(ProductParserImpl.class)
+                && !interfaceClazz.equals(FileReaderServiceImpl.class)
+                && !interfaceClazz.equals(FileReaderService.class)) {
```
Sinelife, line 29: this if statement contains lots of lines which will only grow if we will have more classes in future.

You put this classes in map in findImplementation() method so you can just call findImplementation method to get implementation of your interface that you put in method.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,82 @@
...
+        } catch (NoSuchMethodException | InvocationTargetException
+                | InstantiationException | IllegalAccessException e) {
+            throw new RuntimeException("Instance is not created: " + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
```
Sinelife, line 76: you didnt put ProductService and ProductServiceImpl here

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,82 @@
...
+                && !interfaceClazz.equals(ProductParser.class)
+                && !interfaceClazz.equals(ProductParserImpl.class)
+                && !interfaceClazz.equals(FileReaderServiceImpl.class)
+                && !interfaceClazz.equals(FileReaderService.class)) {
+            throw new RuntimeException("Unsupported class:" + interfaceClazz);
+        }
+        Object clazzImplementationInstance = null;
+        if (interfaceClazz.equals(ProductService.class)) {
+            interfaceClazz = ProductServiceImpl.class;
+        }
```
Sinelife, line 35: You didnt check if class has Component annotation in your Injector.
You should check this. And if class didnt have a Component annotation you shouldn't pass.

Sinelife (changes requested):

src/main/java/mate.academy/lib/Injector.java
Evgen-Krasnenkov: done

src/main/java/mate.academy/lib/Injector.java
Evgen-Krasnenkov: make sense. Thanks. reworked

src/main/java/mate.academy/lib/Injector.java
Evgen-Krasnenkov: I think Component is just for reference. We check fields annotations only. I mean Inject. If I delete @componet form here http://prntscr.com/1thpf6o. Code works.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        Field[] fields = clazz.getDeclaredFields();
+        for (Field field : fields) {
```
Sinelife, line 27: Here before we start for loop we must check if our class has annotation Component. If not we must throw Exception.
 Exercise said.
If Component annotation is missing above the class, we shouldn't be able to create an instance of this class in Injector, and we should throw an exception

Sinelife (changes requested):

src/main/java/mate.academy/lib/Injector.java
Sinelife: Task said us "If Component annotation is missing above the class, we shouldn't be able to create an instance of this class in Injector, and we should throw an exception"

Sinelife (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,79 @@
...
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        if (interfaceClazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("This class doesn't have @Component annotation: "
+                    + interfaceClazz);
+        }
```
MaliukDaria, line 29: Why do we throw exception if annotation is present?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,79 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                | InstantiationException | IllegalAccessException e) {
```
MaliukDaria, line 60: Can we replace this 4 exception with common parent?

MaliukDaria (changes requested):

src/main/java/mate.academy/lib/Injector.java
Evgen-Krasnenkov: done

src/main/java/mate.academy/lib/Injector.java
Evgen-Krasnenkov: sure thing. Done

MaliukDaria (approved):

MaliukDaria (approved):

PR #103, opened 2021-09-26
hw-dependency-injection-rostyslav

RostyslavF5R:


SergiiLonskyi (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+        }
+        if (clazzImplementationInstance == null) {
+            clazzImplementationInstance = createNewInstance(clazz);
+        }
+        return clazzImplementationInstance;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("We cannot create instance of this class. "
```
Max-Pochepets, line 49: of what class?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return constructor.newInstance();
+        } catch (NoSuchMethodException | InvocationTargetException
+                | InstantiationException | IllegalAccessException e) {
+            throw new RuntimeException("Cannot create instance of " + clazz.getName());
```
Max-Pochepets, line 62:
```suggestion
            throw new RuntimeException("Cannot create instance of " + clazz.getName(), e);
```

Max-Pochepets, line 62: there is the same behaviour for all these exceptions. they can be composed into one

Max-Pochepets (changes requested):

Max-Pochepets (approved):

Max-Pochepets (changes requested):
test

Max-Pochepets (approved):

PR #125, opened 2021-09-27
Development

pustoj0:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,60 @@
...
+        if (instances.containsKey(implementationClass)) {
+            return instances.get(implementationClass);
+        }
+        try {
+            Constructor<?> constructor = implementationClass.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(implementationClass, instance);
+            return instance;
+        } catch (InstantiationException | InvocationTargetException
+                | NoSuchMethodException | IllegalAccessException e) {
```
andrii-demchenko, line 55: Let's replace multiple exceptions in the catch block with one common parent exception(checklist, N5)

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,60 @@
...
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.util.HashMap;
+import java.util.Map;
+
 public class Injector {
     private static final Injector injector = new Injector();
 
+    private Map<Class<?>, Object> instances = new HashMap<>();
```
andrii-demchenko, line 12: Empty line
```suggestion
    private static final Injector injector = new Injector();
    private Map<Class<?>, Object> instances = new HashMap<>();
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,60 @@
...
+
     public static Injector getInjector() {
         return injector;
     }
 
-    public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+    public Object getInstance(Class<?> interfaceClass) {
+        Class<?> classImplementation = ImplementationMap.get(interfaceClass);
```
andrii-demchenko, line 19: What if we pass here a class without implementation? As i can see we'll get a null here.

src/main/java/mate.academy/lib/ImplementationMap.java
```diff
@@ -0,0 +1,21 @@
...
+
+public class ImplementationMap {
+    private static final Map<Class<?>, Class<?>> interfaceImplementations
+            = Map.of(FileReaderService.class, FileReaderServiceImpl.class,
+            ProductParser.class, ProductParserImpl.class,
+            ProductService.class, ProductServiceImpl.class);
+
+    public static Class<?> get(Class<?> interfaceClass) {
+        return interfaceImplementations.get(interfaceClass);
+    }
```
andrii-demchenko, line 19: Let's add a "if" clause here, which checks if passed value is interface or not. (In addition to comment in Injector class)

andrii-demchenko (changes requested):

src/main/java/mate.academy/lib/Injector.java
pustoj0: static and nonstatic fields must be separated

src/main/java/mate.academy/lib/Injector.java
pustoj0: Well, I read it, we can do it, no doubt, I just didn't want to, but whatever.

andrii-demchenko (approved):

src/main/java/mate.academy/lib/ImplementationMap.java
```diff
@@ -0,0 +1,24 @@
...
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
+public class ImplementationMap {
+    private static final Map<Class<?>, Class<?>> interfaceImplementations
+            = Map.of(FileReaderService.class, FileReaderServiceImpl.class,
+            ProductParser.class, ProductParserImpl.class,
+            ProductService.class, ProductServiceImpl.class);
+
+    public static Class<?> get(Class<?> interfaceClass) {
```
MaliukDaria, line 17:
```suggestion
    public static Class<?> getImplementation(Class<?> interfaceClass) {
```

MaliukDaria (approved):

PR #126, opened 2021-09-28
Hw jv dependency injection

ivashchenko5:


Yehor-Zapara (changes requested):

Yehor-Zapara (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,71 @@
...
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        Field[] declaredFields = clazz.getDeclaredFields();
+        Object clazzImplementationInstance = createNewInstance(clazz);
+        if (!clazz.isAnnotationPresent(Component.class)) {
```
Max-Pochepets, line 28: you've already created a clazz instance before checking for `@Component`. is it OK?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,71 @@
...
+        if (classInstances.containsKey(clazz)) {
+            return classInstances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            classInstances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                | InstantiationException | IllegalAccessException e) {
```
Max-Pochepets, line 57: here is the same behaviour for all these exceptions. they can be combined into one

Max-Pochepets (changes requested):

Max-Pochepets (approved):

PR #127, opened 2021-09-28
injector was created

Black3942:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,72 @@
...
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | IllegalAccessException
+                | InstantiationException | InvocationTargetException e) {
+            throw new RuntimeException("Can't create new instance of " + clazz.getName());
+        }
```
dogovor, line 61: replace to one parent exception

dogovor (changes requested):

Max-Pochepets (approved):

PR #129, opened 2021-09-29
Develope

forsfakel:
This project implement annotation injector.
Class injector has method:
getInstance - all magic injector is here.
createNewInstance - method to create new object in field
fineImplementation - method to found from map needed class implementation if field is interface and return class object if field is class. And this method run to validate input class and interface.

### DIFF
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -62,7 +62,7 @@ public class Injector {
             return instance;
         } catch (NoSuchMethodException | InvocationTargetException
                 | InstantiationException | IllegalAccessException e) {
-            throw new RuntimeException("Can not create clazz");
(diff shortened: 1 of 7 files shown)


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,83 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
```
RomanDubovskyi, line 63: these exceptions may be replaced with their common parent

RomanDubovskyi (approved):

PR #133, opened 2021-11-19
injector is ready to inject

Andrey-Sidorkin:


src/main/java/mate.academy/service/ProductParser.java
```diff
@@ -11,4 +11,5 @@
      * @return newly create {@link Product} object based on the `productInfo` input.
      */
     Product parse(String productInfo);
+
```
KseniiaMakarova, line 14: redundant empty line

src/main/java/mate.academy/service/impl/FileReaderServiceImpl.java
```diff
@@ -4,9 +4,12 @@
 import java.io.IOException;
 import java.nio.file.Files;
 import java.util.List;
+import mate.academy.lib.Component;
 import mate.academy.service.FileReaderService;
 
+@Component
 public class FileReaderServiceImpl implements FileReaderService {
+
```
KseniiaMakarova, line 12: same

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,84 @@
...
-        return null;
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        Field[] declaredFields = clazz.getDeclaredFields();
+        if (declaredFields.length == 0
+                || Arrays.stream(declaredFields)
+                .filter(f -> f.getType().isPrimitive())
+                .count() == declaredFields.length) {
+            return createNewInstance(clazz);
+        }
```
KseniiaMakarova, line 33: why do you need this?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,84 @@
...
+        for (Field field : declaredFields) {
+            if (!field.getType().isPrimitive()
+                    && !findImplementation(field.getType()).isAnnotationPresent(Component.class)
+            ) {
```
KseniiaMakarova, line 37: you are finding the implementation here for even initialized fields with no `@Inject` annotation? this is not correct, let's remove this as well

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,84 @@
...
+                throw new RuntimeException("No '@Component' Annotation");
+            } else if (!field.getType().isPrimitive()
```
KseniiaMakarova, line 39: else after throw is redundant

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,84 @@
...
+                    && field.isAnnotationPresent(Inject.class)
+                    && findImplementation(field.getType()).isAnnotationPresent(Component.class)) {
```
KseniiaMakarova, line 41: let's check for Component annotation only in 1 place - for example before creating the instance

KseniiaMakarova (changes requested):

src/main/java/mate.academy/lib/Injector.java
Andrey-Sidorkin: It didn't work without it. Objects were null if they hadn't any fields or came without any fields if their fields were primitive. I spent a lot of time to find why and this helped.

src/main/java/mate.academy/lib/Injector.java
Andrey-Sidorkin: just in case. if there is an annotation, but field was primitive. as i wrote above, primitive fields were not initialized without additional code

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                | InstantiationException | IllegalAccessException e) {
```
KseniiaMakarova, line 59: let's try to replace these 4 exceptions with 1

KseniiaMakarova (approved):

PR #172, opened 2021-11-29
dependency injection solution

mykolaskrypka:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+            return instances.get(clazz);
+        }
+        Constructor<?> constructor = null;
+        try {
+            constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (InvocationTargetException | NoSuchMethodException
+                | InstantiationException | IllegalAccessException e) {
```
RomanDubovskyi, line 60: these 4 exceptions can be replaced with their common parent

RomanDubovskyi (approved):

PR #174, opened 2021-12-02
Wrote an assignment jv-dependency-injection.

ASelivonchyk:
- implemented Injector according to the conditions.

### DIFF
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,13 +1,77 @@
 package mate.academy.lib;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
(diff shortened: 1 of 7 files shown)


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+    private Object createNewInstance(Class<?> clazz) {
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
```
RomanDubovskyi, line 60: we can use here their common reflection parent exception

RomanDubovskyi (approved):

PR #176, opened 2021-12-03
Home work

Noct2000:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+
+        Class<?> implClazz = getImpl(interfaceClazz);
+        Object clazzImplementationObject = createNewInstance(implClazz);
+        if (!implClazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Annotation @Component is not present in the class: "
+                    + implClazz.getName());
+        }
```
MaliukDaria, line 30: I think we should first check class for Component annotation and then create it's instance

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+
+        Class<?> implClazz = getImpl(interfaceClazz);
+        Object clazzImplementationObject = createNewInstance(implClazz);
```
MaliukDaria, line 26: I think we should create instance of implClazz only if annotation inject is present, lets move this row in `if (field.isAnnotationPresent(Inject.class)) {`

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+                }
+
+            }
+        }
+        return clazzImplementationObject;
+    }
+
+    private Object createNewInstance(Class<?> implClazz) {
+        if (instances == null) {
+            instances = new HashMap<>();
```
MaliukDaria, line 50: I think it's better to initialize instances Map at the time of declaration (`private final Map<Class<?>, Object> instances = new HashMap<>();`)

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+        try {
+            Constructor<?> constructor = implClazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(implClazz, instance);
+            return instance;
+
+        } catch (NoSuchMethodException
+                | IllegalAccessException
+                | InstantiationException
+                | InvocationTargetException e) {
```
MaliukDaria, line 63: we can replace here these exceptions with their common parent

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+                    + implClazz.getName(), e);
+        }
+    }
+
+    private Class<?> getImpl(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> clazzImplementationMap = new HashMap<>();
+        clazzImplementationMap.put(FileReaderService.class, FileReaderServiceImpl.class);
+        clazzImplementationMap.put(ProductParser.class, ProductParserImpl.class);
+        clazzImplementationMap.put(ProductService.class, ProductServiceImpl.class);
+        return clazzImplementationMap.get(interfaceClazz);
```
MaliukDaria, line 74: what if interfaceClazz is not an interface?

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #184, opened 2022-01-20
Hw 14 solution

djmolot:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,84 @@
...
         return injector;
     }
 
-    public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+    public Object getInstance(Class<?> interfaceClazz) { //interfaceClazz = interfaceName
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Unsupported class is passed" + clazz.getName());
```
MaliukDaria, line 28:
```suggestion
            throw new RuntimeException("Class " + clazz.getName() + " doesn`t have @Component annotation");
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,84 @@
...
+            return instances.get(clazz);
+        }
+        //create a new object
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                | InstantiationException | IllegalAccessException e) {
```
MaliukDaria, line 68: Can we replace these 4 exception with a common parent?

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #185, opened 2022-01-21
Hw custom injector main

maxkmaxk:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,78 @@
...
+        }
+    }
+
+    private Class<?> findImplementationClass(Class<?> interfaceClazz) {
+        if (interfaceClazz.isInterface()) {
+            return classMap.get(interfaceClazz);
+        }
+        if (Modifier.isAbstract(interfaceClazz.getModifiers())) {
+            return classMap.get(interfaceClazz);
+        }
```
MaliukDaria, line 75: Why do we need this?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,78 @@
...
+        if (instances.containsKey(implClass)) {
+            return instances.get(implClass);
+        }
+        try {
+            Constructor<?> constructor = implClass.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(implClass, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                | InstantiationException | IllegalAccessException e) {
```
MaliukDaria, line 64: Can we replace these 4 exception with a common parent?

MaliukDaria (changes requested):

Resci (approved):

PR #197, opened 2022-02-16
imp dependency injection

anyutamysko:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                 | InstantiationException | IllegalAccessException e) {
+            throw new RuntimeException("Cann`t create a new instance of " + clazz.getName(), e);
+        }
+
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
```
LinkClink, line 67: (recomendation) We always creating a map when we calling method findImplementation.
You can use a static block initialisation to create this map.

LinkClink (approved):
Good.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,80 @@
...
+            return instances.get(clazz);
+        }
+        Constructor<?> constructor = null;
+        try {
+            constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                 | InstantiationException | IllegalAccessException e) {
```
KostyaNikiforov, line 68:
```suggestion
        } catch (ReflectiveOperationException e) {
```

KostyaNikiforov (approved):

PR #199, opened 2022-02-18
Ingector class' methods were implemented

Liubov-Chernohor:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object classImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Injection failed,"
+                    + " missing @Component annotation on the class "
+                    + interfaceClazz.getName());
```
MaliukDaria, line 29:
```suggestion
        if (!clazz.isAnnotationPresent(Component.class)) {
            throw new RuntimeException("Injection failed,"
                    + " missing @Component annotation on the class "
                    + clazz.getName());
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | IllegalAccessException
+                | InstantiationException | InvocationTargetException e) {
```
MaliukDaria, line 61: can we replace these 4 exceptions with a common parent?

MaliukDaria (approved):

PR #201, opened 2022-02-20
Jv dependency injection solution

bigmozes:
Implemented injector according to task conditions

### DIFF
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,13 +1,82 @@
 package mate.academy.lib;
 
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.util.HashMap;
+import java.util.Map;
+
 public class Injector {
+    private static final Map<Class<?>, Class<?>> IMPLEMENTATIONS = new HashMap<>();
+    static {
+        IMPLEMENTATIONS.put(ProductService.class, ProductServiceImpl.class);
+        IMPLEMENTATIONS.put(ProductParser.class, ProductParserImpl.class);
+        IMPLEMENTATIONS.put(FileReaderService.class, FileReaderServiceImpl.class);
+    }
+    private static final Map<Class<?>, Object> INSTANCES = new HashMap<>();
     private static final Injector injector = new Injector();
 
     public static Injector getInjector() {
         return injector;
     }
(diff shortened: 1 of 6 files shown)


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,89 @@
...
+import java.util.Map;
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
+    private static final Map<Class<?>, Class<?>> IMPLEMENTATIONS = new HashMap<>();
```
MaliukDaria, line 16:
```suggestion
    private static final Map<Class<?>, Class<?>> implementations = new HashMap<>();
```

https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,89 @@
...
+
+    static {
+        IMPLEMENTATIONS.put(ProductService.class, ProductServiceImpl.class);
+        IMPLEMENTATIONS.put(ProductParser.class, ProductParserImpl.class);
+        IMPLEMENTATIONS.put(FileReaderService.class, FileReaderServiceImpl.class);
+    }
+
+    private static final Map<Class<?>, Object> INSTANCES = new HashMap<>();
```
MaliukDaria, line 24:
```suggestion
    private static final Map<Class<?>, Object> instances = new HashMap<>();
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,89 @@
...
+        if (INSTANCES.containsKey(implementationClass)) {
+            return INSTANCES.get(implementationClass);
+        }
+        try {
+            Constructor<?> constructor = implementationClass.getConstructor();
+            Object instance = constructor.newInstance();
+            INSTANCES.put(implementationClass, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                | InstantiationException | IllegalAccessException e) {
```
MaliukDaria, line 66: can we replace this 4 exceptions with common parent?

MaliukDaria (changes requested):

src/main/java/mate.academy/lib/Injector.java
bigmozes: Right, have to read out task conditions more precisely )

src/main/java/mate.academy/lib/Injector.java
bigmozes: Yeah, sure.

src/main/java/mate.academy/lib/Injector.java
bigmozes: Got it.

MaliukDaria (approved):

PR #205, opened 2022-02-23
implemented Injector

OlehMytiai:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,78 @@
...
+        if (components.containsKey(clazz)) {
+            return components.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            components.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InstantiationException
+                | IllegalAccessException | InvocationTargetException e) {
```
DaniilOpryshko, line 62: Here we can replace multiple exceptions in the catch block with one common parent exception

DaniilOpryshko (approved):

KostyaNikiforov (approved):

PR #206, opened 2022-02-23
was implemented injector

IgorZima90:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("in class" + clazz + " does not exists annotation "
```
MaliukDaria, line 27:
```suggestion
            throw new RuntimeException("in class" + clazz.getName() + " does not have annotation "
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                | InstantiationException | IllegalAccessException e) {
```
MaliukDaria, line 60: can we replace this exceptions with common patent?

MaliukDaria (approved):

PR #208, opened 2022-03-05
1st implementation of jv-dependency-injection task

OleksiyTitov:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Injection failed, missing annotation @Component: "
+                    + interfaceClazz.getName());
+        }
```
MaliukDaria, line 29:
```suggestion
        if (!clazz.isAnnotationPresent(Component.class)) {
            throw new RuntimeException("Injection failed, missing annotation @Component: "
                    + clazz.getName());
        }
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+        Field[] declaredFields = interfaceClazz.getFields();
```
MaliukDaria, line 30:
```suggestion
        Field[] declaredFields = clazz.getDeclaredFields();
```

could interface have fields? What is the difference between getFields() and getDeclaredFields() ?

MaliukDaria, line 30:
```suggestion
        Field[] declaredFields = clazz.getFields();
```

could interface have fields?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InstantiationException
+                | IllegalAccessException | InvocationTargetException e) {
```
MaliukDaria, line 60: could we combine these 4 exceptions?

MaliukDaria, line 60: could we combine these 4 exceptions?

MaliukDaria (changes requested):

MaliukDaria (approved):
👍🏻

PR #213, opened 2022-03-14
implemented an injector that uses @Component and @Inject annotations …

Viktoriia-Zinchenko:
…to initialize and return Object instances


DenTima (approved):
LGTM🔥

KostyaNikiforov (approved):

PR #216, opened 2022-03-14
DI task done

o240789:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+    private Object createNewInstance(Class<?> clazz) {
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
```
antonignatenko, line 58: here you can replace all of these exceptions with one

antonignatenko:
```suggestion
        } catch (ReflectiveOperationException e) {
            throw new RuntimeException("Can't create new instance of " + clazz.getName());
```

antonignatenko (changes requested):

src/main/java/mate.academy/Main.java
```diff
@@ -4,13 +4,14 @@
...
+import mate.academy.service.impl.ProductServiceImpl;
 
 public class Main {
 
     public static void main(String[] args) {
-        // Please test your Injector here. Feel free to push this class as a part of your solution
         Injector injector = Injector.getInjector();
-        ProductService productService = null;
+        ProductService productService = (ProductService)
+                injector.getInstance(ProductServiceImpl.class);
```
KostyaNikiforov, line 14:
```suggestion
                injector.getInstance(ProductService.class);
```

KostyaNikiforov (approved):

PR #217, opened 2022-03-15
Implemented Injector, inserted annotations

NazariiYatsiuk:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+            return instances.get(clazz);
+        }
+        Constructor<?> constructor;
+        try {
+            constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | IllegalAccessException
+                | InstantiationException | InvocationTargetException e) {
```
DaniilOpryshko, line 62: U can replace multiple exceptions in catch block with one common exception

DaniilOpryshko (changes requested):

KostyaNikiforov (approved):

PR #223, opened 2022-04-03
implemented Injector

tanya-logatskaya:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Injection failed, missing "
+                    + "@Component annotation on the class "
+                    + interfaceClazz.getName());
+        }
+        Field[] declaredFields = interfaceClazz.getDeclaredFields();
```
Sofasmile, line 31: should we get fields from the interface?

Sofasmile (changes requested):
let's fix main method

Sofasmile (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create new instance of " + clazz.getName());
```
Sofasmile, line 61:
```suggestion
            throw new RuntimeException("Can't create new instance of " + clazz.getName(), e);
```

PR #224, opened 2022-04-04
solved jv-dependency-injection task

Angela1608:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                | InstantiationException | IllegalAccessException e) {
```
MaliukDaria, line 61: Could we replace these 4 exceptions with a common parent?

MaliukDaria (changes requested):

MaliukDaria (approved):
GJ🔥

PR #229, opened 2022-04-21
Implemented injector using annotations

Bukozdem:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+    private Object createNewInstance(Class<?> clazz) {
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Object instance = clazz.getConstructor().newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                | InstantiationException | IllegalAccessException e) {
```
MaliukDaria, line 55: can we replace these 4 exceptions with a common parent?

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #232, opened 2022-04-30
- implemented injector class with annotations @Injector and @Component

MarkShapka:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,71 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | IllegalAccessException
+                | InstantiationException | InvocationTargetException e) {
```
MaliukDaria, line 56: can we replace these exceptions with common parent?

MaliukDaria (changes requested):
Task: "If Component annotation is missing above the class, we shouldn't be able to create an instance of this class in Injector, and we should throw an exception" -> I don't see that part

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Can't found implementation the interface: "
+                    + interfaceClazz);
```
MaliukDaria, line 27:
```suggestion
            throw new RuntimeException("There is no @Component annotation over class implementation "
                    + interfaceClazz + " of interface " + interfaceClazz);
```

MaliukDaria (approved):

PR #233, opened 2022-05-02
jv_homework_dependency-injector

etezaadi:
implemented injector

### DIFF
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,13 +1,76 @@
 package mate.academy.lib;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.util.HashMap;
+import java.util.Map;
+import mate.academy.service.FileReaderService;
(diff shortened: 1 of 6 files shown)


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = clazz.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException
+                | InstantiationException
+                | IllegalAccessException e) {
```
MaliukDaria, line 60: could we use one common parent for these exceptions?

MaliukDaria (changes requested):

src/main/java/mate.academy/lib/Injector.java
etezaadi: @MaliukDaria sure, we can :)

MaliukDaria (approved):
👍🏻

PR #235, opened 2022-05-03
Injector method was fulfilled

Oleksii1412:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+        if (instances.containsKey(clas)) {
+            return instances.get(clas);
+        }
+        try {
+            Constructor<?> constructor = clas.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clas,instance);
+            return instance;
+        } catch (InvocationTargetException | InstantiationException
+                | IllegalAccessException | NoSuchMethodException e) {
```
MaliukDaria, line 57: can we replace these 4 exceptions with common parent?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+        interfaceImplementation.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementation.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementation.put(ProductService.class, ProductServiceImpl.class);
+        if (interfaceClass.isInterface()) {
+            return interfaceImplementation.get(interfaceClass);
+        }
+        return interfaceImplementation.keySet()
+                .stream()
+                .filter(c -> c.isAnnotationPresent(Component.class))
+                .findFirst().orElseThrow(() -> new RuntimeException("Unsupported class is passed"));
```
MaliukDaria, line 73: we need to check class implementation for @Component annotation, not interface

MaliukDaria (changes requested):

src/main/java/mate.academy/lib/Injector.java
Oleksii1412: thanks for your advice))

MaliukDaria (approved):

PR #260, opened 2022-05-12
Dependency injection task

DenysShl:


src/main/java/mate.academy/service/impl/ProductServiceImpl.java
```diff
@@ -2,17 +2,27 @@
...
+    @Inject
+    private ProductParser productParser;
 
     @Override
     public List<Product> getAllFromFile(String filePath) {
+        if (fileReaderService == null || productParser == null) {
+            throw new RuntimeException("Can`t get all product because this "
+                    + productParser + " or " + fileReaderService
+                    + " is null");
+        }
```
Nick97-git, line 25: did you need to write this code here according to the task?

Nick97-git (changes requested):

Nick97-git (approved):

PR #261, opened 2022-05-12
 jv-dependency-injection, could you check it, please

lutskvadym:


src/main/java/mate.academy/model/Product.java
```diff
@@ -9,10 +9,6 @@ public class Product {
     private String description;
     private BigDecimal price;
 
-    public Long getId() {
-        return id;
-    }
-
```
Nick97-git, line 15: you didn't need to remove these methods according to the task, return them

Nick97-git (changes requested):

Nick97-git (approved):

PR #263, opened 2022-05-17
created dependency injection

DLyudmila:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        Object classImplementationInstance = null;
+        if (clazz.isAnnotationPresent(Component.class)) {
```
MaliukDaria, line 26:
```suggestion
        if (!clazz.isAnnotationPresent(Component.class)) {
        	throw new RuntimeException("Cant create instance of class " + clazz.getSimpleName() 
        		+ " . Class does not have @Component annotation" );
        }
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                | InstantiationException | IllegalAccessException e) {
```
MaliukDaria, line 62: could we replace this 4 exceptions with common parent

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #266, opened 2022-05-27
added solution for jv-dependency-injection task

dmnichz:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+            throw new RuntimeException("Annotation 'Component' is missing in "
+                    + clazz.getName());
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                | InstantiationException | IllegalAccessException e) {
```
MaliukDaria, line 60: could we replace these 4 exceptions with common parent?

MaliukDaria (approved):

src/main/java/mate.academy/lib/Injector.java
dmnichz: yes, I think we can replace these exceptions with ReflectiveOperationException.

PR #289, opened 2022-06-20
 Jv-dependency-injection-maxlut

maxlutovinov:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,78 @@
...
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.ServiceLoader;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private static final Map<Class<?>, Object> instancesMap = new HashMap<>();
+    private static final Map<Class<?>, Class<?>> interfacesImplementationClassMap = new HashMap<>();
```
Nick97-git, line 12:
```suggestion
    private static final Map<Class<?>, Class<?>> interfaceImplementationsMap = new HashMap<>();
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,78 @@
...
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Class<?> implementationClazz = findImplementationClass(interfaceClazz);
+        /* It seems to be better to move this check to the 'findImplementationClass()' method
+        before filling the 'interfacesImplementationClassMap', but then one test fails */
```
Nick97-git, line 21: remove comments

Nick97-git (changes requested):
left comments + what these files are doing here?
Screenshot 2022-06-21 at 08 26 48

src/main/java/mate.academy/lib/Injector.java
maxlutovinov: Removed

maxlutovinov:
> left comments + what these files are doing here? Screenshot 2022-06-21 at 08 26 48

These files are needed for the ServiceLoader to work. This way I find the implementations of interfaces.

Nick97-git (changes requested):
then rewrite your logic, so you won't need to add these files to PR, it's enough to add to map directly all implementations

maxlutovinov:
> then rewrite your logic, so you won't need to add these files to PR, it's enough to add to map directly all implementations

Could you explain why you don't like the way of finding  the implementations with ServiceLoader? In the description of this task, there are no restrictions on how I should fill the map. In addition, in the training videos on this topic, the mentor several times focuses on the fact that manual filling of the map is usually not used, instead the implementations map is built programmatically.

Nick97-git (approved):

PR #295, opened 2022-06-26
jv-dependency-injection implemented

drapail:
First implementation without mentor's checks.

### DIFF
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -20,7 +20,7 @@ public class Injector {
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        if (!interfaceClazz.isAnnotationPresent(Component.class)) {
+        if (!interfaceClazz.isAnnotationPresent(Component.class) && !interfaceClazz.isInterface()) {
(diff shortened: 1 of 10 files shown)


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        if (!interfaceClazz.isAnnotationPresent(Component.class) && !interfaceClazz.isInterface()) {
+            throw new RuntimeException("We can't create instance of " + interfaceClazz
+                    + " through Injector. "
+                    + interfaceClazz + " class has no @Component annotation");
+        }
+        Class<?> clazz = findImplementation(interfaceClazz);
```
MaliukDaria, line 28:
```suggestion
                Class<?> clazz = findImplementation(interfaceClazz);
        if (!clazz.isAnnotationPresent(Component.class) ) {
            throw new RuntimeException(clazz + " class has no @Component annotation");
        }
```

MaliukDaria (changes requested):
git@github.com:drapail/jv-dependency-injection.git

PR #297, opened 2022-06-28
Injector was created

YaroslavLa:


DimaPodakin (approved):
Good job_))

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can`t create a new instance of " + clazz.getName());
+        }
```
Noct2000, line 61: there too
```suggestion
            return instance;
        } catch (ReflectiveOperationException e) {
            throw new RuntimeException("Can`t create a new instance of " + clazz.getName(), e);
        }
```

Noct2000 (changes requested):

PR #303, opened 2022-06-29
Hw dependency injection

kate-chern:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+        }
+        Object clazzImplementationInsctance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Injection failed, @Component absence on the class "
+                    + clazz.getName());
+        }
+        Field[] declaredFields = interfaceClazz.getDeclaredFields();
+        for (Field field: declaredFields
+             ) {
```
MaliukDaria, line 34:
```suggestion
        for (Field field: declaredFields) {
```

src/main/java/mate.academy/service/FileReaderService.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 
+@Component
```
MaliukDaria, line 6: we should place @Component annotation only over classes - implementations

MaliukDaria (changes requested):

MaliukDaria (changes requested):
comments from previous review not fixed

MaliukDaria (approved):

PR #306, opened 2022-06-29
Injector along with annotations was implemented

olyaklipa:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+            throw new RuntimeException("Injection failed, missing @Component annotaion on "
+                    + clazz.getName() + " class");
+        }
+        Field[] declaredFields = clazz.getDeclaredFields();
+        for (Field field : declaredFields) {
+            if (field.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = getInstance(field.getType());
+                clazzImplementationInstance = createNewInstance(clazz);
+                try {
+                    field.setAccessible(true);
```
oleksii-kordysh, line 35: We could move this line before try block, but in your case, I think it's ok too.

oleksii-kordysh (approved):

src/main/java/mate.academy/Main.java
```diff
@@ -10,7 +10,7 @@ public class Main {
     public static void main(String[] args) {
         // Please test your Injector here. Feel free to push this class as a part of your solution
```
Noct2000, line 11: remove comment

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
```
Noct2000, line 70:
```suggestion
        try {
            Constructor<?> constructor = clazz.getConstructor();
            Object instance = constructor.newInstance();
            instances.put(clazz, instance);
            return instance;
        } catch (ReflectiveOperationException e) {
            throw new RuntimeException("Can't create a new instance of " + clazz.getName(), e);
```

Noct2000 (changes requested):

PR #308, opened 2022-07-01
Injector implemented and annotations added to classes

Illinois357:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,12 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Component {
 
 }
```
zim10093, line 12: Empty line, I prefer this format, but I don`t know if it correct
```suggestion
public @interface Component {}
```

src/main/java/mate.academy/service/ProductParser.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
```
zim10093, line 6: Same

src/main/java/mate.academy/service/ProductService.java
```diff
@@ -1,8 +1,10 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
```
zim10093, line 7: Same

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,86 @@
...
+        if (INSTANCES.containsKey(clazz)) {
+            return INSTANCES.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            INSTANCES.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Couldn't create an Instance of class: " + clazz.getName());
```
zim10093, line 67:
```suggestion
            throw new RuntimeException("Couldn't create an instance of class: " + clazz.getName());
```

zim10093 (changes requested):

src/main/java/mate.academy/lib/Component.java
Illinois357: Let's see what mentor would say

zim10093 (approved):

src/main/java/mate.academy/lib/Component.java
dSkarzh: both are fine as long as you stay consistent with it

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,79 @@
...
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> clazz) {
+        if (implementations.containsKey(clazz)) {
+            return implementations.get(clazz);
+        }
+        if (clazz.isInterface()) {
+            throw new RuntimeException("unknown class: " + clazz + " for injector");
+        }
```
dSkarzh, line 76: should we throw an exception if clazz is interface? maybe it would be better to return it's implementation
and what should we do when we clazz is regular class?

dSkarzh (changes requested):

src/main/java/mate.academy/lib/Injector.java
Illinois357: We throw an exeption if we don't have such an interface in a map

Illinois357: if it is a regular class we return it

PR #312, opened 2022-07-14
Class Injector was implemented.

losnazar:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,79 @@
...
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        Field[] declaredFields = clazz.getDeclaredFields();
+        for (Field field : declaredFields) {
+            if (field.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = getInstance(field.getType());
+                if (clazz.isAnnotationPresent(Component.class)) {
```
MaliukDaria, line 29: do we need to check it in a loop? Could we move this check after line 24? And lets implement it like
```
 throw new RuntimeException(....
 }
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,79 @@
...
+                                + clazz.getName() + ". Field; " + field.getName() + ".");
+                    }
+                } else {
+                    throw new RuntimeException("Injection failed, "
+                            + "missing @Component annotation on the class " + clazz.getName());
+                }
+            }
+        }
+        if (clazzImplementationInstance == null) {
+            if (clazz.isAnnotationPresent(Component.class)) {
```
MaliukDaria, line 45: why do we need to check it twice?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,79 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object newInstance = constructor.newInstance();
+            instances.put(clazz, newInstance);
+            return newInstance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create new instance of " + clazz.getName());
```
MaliukDaria, line 65:
```suggestion
            throw new RuntimeException("Can't create new instance of " + clazz.getName(), e);
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,79 @@
...
+            if (field.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = getInstance(field.getType());
+                if (clazz.isAnnotationPresent(Component.class)) {
+                    clazzImplementationInstance = createNewInstance(clazz);
+                    field.setAccessible(true);
+                    try {
+                        field.set(clazzImplementationInstance, fieldInstance);
+                    } catch (IllegalAccessException e) {
+                        throw new RuntimeException("Can't initialize field value. Class; "
+                                + clazz.getName() + ". Field; " + field.getName() + ".");
```
MaliukDaria, line 36:
```suggestion
                                + clazz.getName() + ". Field; " + field.getName() + ".", e);
```

MaliukDaria (approved):

PR #315, opened 2022-08-05
Hw dependency injection

savchenkoarts:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,81 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | IllegalAccessException
+                 | InstantiationException | InvocationTargetException e) {
```
Nick97-git, line 66: they have a common parent, let's use it instead of them

Nick97-git (changes requested):

Nick97-git (approved):

PR #318, opened 2022-08-07
implementation of task conditions

maksym-bakunov:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                 | InstantiationException | IllegalAccessException e) {
```
MaliukDaria, line 60: Could we replace these 4 exceptions with common parent?

MaliukDaria (approved):

src/main/java/mate.academy/lib/Injector.java
maksym-bakunov: ReflectiveOperationException. Is it right?

PR #320, opened 2022-08-10
completed task jv-dependency-injection

Oleg-pyt:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,62 @@
...
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        if (!interfaceClazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Class " + interfaceClazz
+                    + " can`t have annotation Component");
+        }
```
Sofasmile, line 26: do we need this check?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,62 @@
...
+    private Object createNewInstance(Class<?> clazz) {
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Object instance = clazz.getConstructor().newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                 | InstantiationException | IllegalAccessException e) {
```
Sofasmile, line 50: they have a common parent, so let's use it here instead of 4 of them

src/main/java/mate.academy/service/FileReaderService.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 
+@Component
 public interface FileReaderService {
```
Sofasmile, line 7: we should put annotation only above the class

Sofasmile (changes requested):

Sofasmile (approved):

PR #332, opened 2022-08-29
create implementation of inject class, use Annotation Component

verbanov:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
Oleksandr-Tovstyi, line 6: You can add @Target(ElementType.TYPE) here. TYPE means that it used with classes. But it's not necessery.

verbanov: Thank you Sasha, I'll add it

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+            }
+        }
+        if (classImplementationInstance == null) {
+            classImplementationInstance = createNewInstance(clazz);
+        }
+        return classImplementationInstance;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
+        if (clazz.isAnnotationPresent(Component.class)) {
```
dSkarzh, line 46: move this check to the top of `getInstance()`

dSkarzh (changes requested):

dSkarzh (approved):

PR #336, opened 2022-08-31
Implement Injector class methods: getInstance(), createNewInstance(),…

voitalik:
… findImplementation()

### DIFF
--- a/src/main/java/mate.academy/lib/Component.java
+++ b/src/main/java/mate.academy/lib/Component.java
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Component {
 
 }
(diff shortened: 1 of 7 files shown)


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        if (clazz.isAnnotationPresent(Component.class)) {
```
kozhukhovsky, line 25: It would be better to check first for annotation present and throw exception, if not.
And under the first if block, write other code, like:
   throw exception
}
....
....

src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Component {
```
kozhukhovsky, line 7: You can add `@Target(ElementType.TYPE)` here

kozhukhovsky (changes requested):

aaananas (approved):

PR #340, opened 2022-09-02
implemented Injector class which uses @Component and @Inject annotations

GolDenik:


JustElvis (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object currentInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Injection failed. @Component annotation is missed in "
+                    + interfaceClazz.getName());
```
Sofasmile, line 27:
```suggestion
                    + clazz.getName());
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+        }
+        Field[] fields = interfaceClazz.getFields();
```
Sofasmile, line 29: what is the difference between getFields() and getDeclaredFields()?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+        for (Field field : fields) {
+            if (field.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = injector.getInstance(field.getType());
+
```
Sofasmile, line 33: remove empty line

Sofasmile (changes requested):
let's also modify the main() method

src/main/java/mate.academy/lib/Injector.java
GolDenik: getFields() can get fields from SuperClass as well, right ?

GolDenik: No , getDeclaredFields can have acces to private fields , right ?

src/main/java/mate.academy/Main.java
```diff
@@ -4,13 +4,15 @@
...
+import mate.academy.service.impl.ProductServiceImpl;
 
 public class Main {
 
     public static void main(String[] args) {
         // Please test your Injector here. Feel free to push this class as a part of your solution
         Injector injector = Injector.getInjector();
-        ProductService productService = null;
+        ProductService productService =
+                (ProductServiceImpl) injector.getInstance(ProductService.class);
```
Sofasmile, line 15:
```suggestion
                (ProductService) injector.getInstance(ProductService.class);
```

Sofasmile (changes requested):

src/main/java/mate.academy/lib/Injector.java
Sofasmile: we can access to all fields

Sofasmile (approved):

PR #343, opened 2022-09-03
Added my solution.

vadimkik2:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+    private Object createNewInstance(Class<?> clazz) {
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | IllegalAccessException
```
max-bolotin, line 70: As the checklist says, it is better to replace many exceptions that have a common parent with a general parental exception:
Good example: 

max-bolotin (changes requested):

max-bolotin (approved):
Great!

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
```
Noct2000, line 70: add root cause exception

Noct2000 (changes requested):

Noct2000 (approved):

PR #345, opened 2022-09-04
Injector was created

mikhail-maiorov:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
verbanov, line 6: it is not nesessary but you can put @Target (Element Type) parametres here

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+            clazzImplementationInstance = createNewInstance(clazz);
+        }
+        return clazzImplementationInstance;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+
```
verbanov, line 50: empty line

src/main/java/mate.academy/service/FileReaderService.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 
+@Component
```
verbanov, line 6: you should use @Component annotation only over implementations of interfaces, not over interfaces

src/main/java/mate.academy/service/ProductParser.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
```
verbanov, line 6: you should use https://github.com/component annotation only over implementations of interfaces, not over interfaces

src/main/java/mate.academy/service/ProductService.java
```diff
@@ -1,8 +1,10 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
```
verbanov, line 7: and here

verbanov (changes requested):

verbanov (approved):

Sofasmile (approved):

PR #377, opened 2022-09-05
Add implementation of Injector

LubusUzh:


Alexhqolma (approved):

src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Component {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
```
Nesterenko-Andrey, line 6: you cal also add
`@Target(ElementType.TYPE)`

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Inject {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
```
Nesterenko-Andrey, line 6: same here
`@Target(ElementType.FIELD)`

Nesterenko-Andrey (changes requested):

src/main/java/mate.academy/lib/Inject.java
LubusUzh: I'll try it, but it doesn't work (

Nesterenko-Andrey (approved):

Sofasmile (approved):

PR #383, opened 2022-09-13
Jv hv dependency injection

DooMLeoN:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,84 @@
...
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz,instance);
+            return instances.get(clazz);
+        } catch (NoSuchMethodException
+                | InvocationTargetException
+                | InstantiationException
+                | IllegalAccessException e) {
```
aaananas, line 64: add parent exception

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,84 @@
...
+            throw new RuntimeException("Can not create a new instance: "
+                    + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceAndImpl = new HashMap<>();
+        interfaceAndImpl.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceAndImpl.put(ProductService.class, ProductServiceImpl.class);
+        interfaceAndImpl.put(ProductParser.class, ProductParserImpl.class);
```
aaananas, line 74: you can make this map global and only get value from it
in this realization a new map would be created each time you invoke this method

aaananas (changes requested):

aaananas (approved):

PR #385, opened 2022-09-17
Implemented Injector

dmytrovb02:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,72 @@
...
+        if (instance == null) {
+            return createInstance(realization);
+        }
+        return instance;
+    }
+
+    private Class<?> findRealization(Class<?> interfaceClazz) {
+        interfaceImpl.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImpl.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImpl.put(ProductService.class, ProductServiceImpl.class);
```
aaananas, line 52: initialize the map with those values in static block)

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,72 @@
...
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private final Map<Class<?>, Object> instances = new HashMap<>();
+    private final Map<Class<?>, Class<?>> interfaceImpl = new HashMap<>();
```
```suggestion
    private final Map<Class<?>, Class<?>> implementations = new HashMap<>();
```

aaananas (changes requested):

kozhukhovsky (approved):

PR #393, opened 2022-09-28
Impl Injector and add custom annotations

IvanAldokhin:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,87 @@
...
+        isAnnotationComponentPresent(findClass);
+        return findClass;
+    }
+
+    private Map<Class<?>,Class<?>> getMapInterfaceAndImplement() {
+        Map<Class<?>,Class<?>> interfaceMapImpl = new HashMap<>();
+        interfaceMapImpl.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceMapImpl.put(ProductParser.class, ProductParserImpl.class);
+        interfaceMapImpl.put(ProductService.class, ProductServiceImpl.class);
+        return interfaceMapImpl;
```
aaananas, line 77: you can make this map global, this way you wont create a new map each time findImplement method will be invoked

aaananas (approved):

PR #394, opened 2022-09-29
hw done

oleksandr-kravchenko:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz,instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                 | InstantiationException | IllegalAccessException e) {
```
aaananas, line 60:
```suggestion
        } catch (ReflectiveOperationException e) {
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+            throw new RuntimeException("Cant create a new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>,Class<?>> interfaceImplementation = new HashMap<>();
+        interfaceImplementation.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementation.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementation.put(ProductService.class, ProductServiceImpl.class);
```
aaananas, line 69: you  can make this map global, and initialize in the initialization block,
because the same map will be created every time this method is invoked

aaananas (changes requested):
Great job

PR #395, opened 2022-10-02
implemented Injector class, added Inject and Component annotations

NataliaKurhanska:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+            clazzImplementationInstance = createNewInstance(clazz);
+        }
+        return clazzImplementationInstance;
+    }
+
+    private static Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
```
aaananas, line 54: just a suggestion, you can make this map global, and initialize in initialization block
this way it will be created only once

aaananas (approved):

PR #396, opened 2022-10-02
Hw

Oldudka:


src/main/java/mate.academy/Main.java
```diff
@@ -4,13 +4,15 @@
...
+import mate.academy.service.impl.ProductServiceImpl;
 
 public class Main {
 
     public static void main(String[] args) {
         // Please test your Injector here. Feel free to push this class as a part of your solution
         Injector injector = Injector.getInjector();
-        ProductService productService = null;
+        ProductService productService = (ProductService) injector
+                .getInstance(ProductServiceImpl.class);
```
kozhukhovsky, line 15:
```suggestion
        ProductService productService = (ProductService) injector
                .getInstance(ProductService.class);
```

kozhukhovsky: And you need add annotation "@Component" to interfaces

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
```
kozhukhovsky, line 67: you can initialize in initialization block
this way it will be created only once

kozhukhovsky (changes requested):

aaananas (approved):

PR #400, opened 2022-10-08
added solution

roman680:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+        if (classImplementationInstance == null) {
+            classImplementationInstance = createNewInstance(clazz);
+        }
+        return classImplementationInstance;
+    }
+
+    private static Class<?> findImplementation(Class<?> interfaceClass) {
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
```
aaananas, line 53: lets initialize the map once in the top (you can use Map.of(key value, key, value...))
instead of every time this method is invoked

aaananas (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
         return injector;
     }
 
-    public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+    public Object getInstance(Class<?> interfaceClass) {
+        if (!interfaceClass.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException(interfaceClass.getName()
+                    + " does not marked with Component annotation");
+        }
```
MaliukDaria, line 33: Component annotation should be present over class-implementation, not interface

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create an instance of " + clazz.getName());
```
MaliukDaria, line 73:
```suggestion
            throw new RuntimeException("Can't create an instance of " + clazz.getName(), e);
```

src/main/java/mate.academy/service/ProductService.java
```diff
@@ -1,8 +1,10 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
```
MaliukDaria, line 7: we dont need this annotation here

src/main/java/mate.academy/service/ProductParser.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
```
MaliukDaria, line 6: we dont need this annotation here, it should be over class-implementation of this interface

MaliukDaria (changes requested):

src/main/java/mate.academy/service/ProductService.java
roman680: but in test we need this

kateryna-deshchinska (approved):

PR #402, opened 2022-10-11
Solution was added

Kostiantyn-Sacheiev:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (InstantiationException | NoSuchMethodException
+                 | InvocationTargetException | IllegalAccessException e) {
```
kateryna-deshchinska, line 58: It is better to replace many exceptions that have a common parent with a general parental exception.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+                    field.set(clazzImplInstance, fieldInstance);
+                } catch (IllegalAccessException e) {
+                    throw new RuntimeException("Can't initialize field value. "
+                            + "Class: " + clazz.getName() + ". Field: " + field.getName());
+                }
+            }
+        }
+        if (!clazzImplInstance.getClass().isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("You pass unsupported class."
+                    + " Please check annotation");
```
kateryna-deshchinska, line 43: Let’s check instances map before new instance creation.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+                try {
+                    field.setAccessible(true);
+                    field.set(clazzImplInstance, fieldInstance);
+                } catch (IllegalAccessException e) {
+                    throw new RuntimeException("Can't initialize field value. "
+                            + "Class: " + clazz.getName() + ". Field: " + field.getName());
+                }
+            }
+        }
+        if (!clazzImplInstance.getClass().isAnnotationPresent(Component.class)) {
```
kateryna-deshchinska, line 41:
```suggestion
        if (!clazz.isAnnotationPresent(Component.class)) {
```

src/main/java/mate.academy/lib/Injector.java
kateryna-deshchinska, line 43: When throwing an exception add an informative message to it.

kateryna-deshchinska (changes requested):

src/main/java/mate.academy/lib/Injector.java
Kostiantyn-Sacheiev: image
Isn't this a check?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
+        interfaceImplementations.put(FileReaderService.class,
+                FileReaderServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class,
+                ProductParserImpl.class);
+        interfaceImplementations.put(ProductService.class,
+                ProductServiceImpl.class);
```
aaananas, line 68: you can make this map global, to only create it once and just invoke in this method
`Map.of(key, value, key, value)

aaananas (approved):

PR #403, opened 2022-10-12
Introduced injector implementation according to the task description.

vasiliy-sheretko:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,82 @@
...
+            }
+        }
+        return clazzImplementationInstance;
+    }
+
+    private Object createNewInstance(Class<?> clazz) throws Exception {
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        if (clazz.isAnnotationPresent(Component.class)) {
```
kateryna-deshchinska, line 59: add this check in getInstance() method before creation instance, and throw RuntimeException with message if it is false. And after that you can delete redundant try-catch block inside getInstance method

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,82 @@
...
+        Class<?> clazz = findImplementation(interfaceClazz);
+        Field[] declaredFields = clazz.getDeclaredFields();
+        for (Field field: declaredFields) {
+            if (field.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = getInstance(field.getType());
+                try {
+                    clazzImplementationInstance = createNewInstance(clazz);
+                } catch (Exception e) {
+                    throw new RuntimeException("Injection failed, missing @Component annotaion "
+                            + "on the class " + clazz.getName(), e);
```
kateryna-deshchinska, line 33: Its redundant try catch block here

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,82 @@
...
+                }
+            }
+        }
+        if (clazzImplementationInstance == null) {
+            try {
+                clazzImplementationInstance = createNewInstance(clazz);
+            } catch (Exception e) {
+                throw new RuntimeException("Injection failed, missing @Component annotaion "
+                        + "on the class " + clazz.getName(), e);
+            }
```
kateryna-deshchinska, line 50: Its redundant try catch block

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,82 @@
...
+        }
+        return clazzImplementationInstance;
+    }
+
+    private Object createNewInstance(Class<?> clazz) throws Exception {
```
kateryna-deshchinska, line 55:
```suggestion
    private Object createNewInstance(Class<?> clazz) {
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,82 @@
...
+            try {
+                Constructor<?> constructor = clazz.getConstructor();
+                Object instance = constructor.newInstance();
+                instances.put(clazz, instance);
+                return instance;
+            } catch (ReflectiveOperationException e) {
+                throw new RuntimeException("Can't create a new instance of " + clazz.getName());
+            }
+        }
+        throw new Exception();
```
kateryna-deshchinska, line 69:
```suggestion
```

src/test/java/mate/academy/lib/InjectorTest.java
```diff
@@ -14,7 +14,7 @@
 import java.util.Arrays;
 
 class InjectorTest {
-    private Injector injector = Injector.getInjector();
+    private final Injector injector = Injector.getInjector();
```
kateryna-deshchinska, line 17: Don't need to change test class

kateryna-deshchinska (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,80 @@
...
+        Field[] declaredFields = clazz.getDeclaredFields();
+        for (Field field: declaredFields) {
+            if (field.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = getInstance(field.getType());
+                if (clazz.isAnnotationPresent(Component.class)) {
+                    clazzImplementationInstance = createNewInstance(clazz);
+                } else {
+                    throw new RuntimeException("Injection failed, missing @Component annotaion "
+                            + "on the class " + clazz.getName());
+                }
```
zachar-chorny, line 34: Do this check at the beginning of the method and if it is false throw an exception. After all, we don't need it in every iteration. Just like the call of createNewInstance(clazz) method does not make sense to add to the body of if

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,80 @@
...
+            }
+        }
+        if (clazzImplementationInstance == null) {
+            if (clazz.isAnnotationPresent(Component.class)) {
+                clazzImplementationInstance = createNewInstance(clazz);
+            } else {
+                throw new RuntimeException("Injection failed, missing @Component annotaion "
+                        + "on the class " + clazz.getName());
+            }
```
zachar-chorny, line 50: This is completely unnecessary here, again checking at the beginning of the method will solve everything
```suggestion
            clazzImplementationInstance = createNewInstance(clazz);
```

zachar-chorny (changes requested):

kateryna-deshchinska (approved):

PR #406, opened 2022-10-22
 task solution

shlop-shlop:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of: " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
```
aaananas, line 68: you can make this map global, this way it will be created just once, instead of each time this method is invoked

aaananas (approved):

PR #408, opened 2022-10-27
implemented Injector class

nazariipertsak:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,72 @@
...
+        return clazzImplementationInstance;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
```
CyberDogFK, line 54: I think we get Object from clazz in one line

CyberDogFK (changes requested):

CyberDogFK (approved):
LGTM

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
ChumakAn, line 6: you can add target here too

ChumakAn (changes requested):
Good! Minor issue

ChumakAn (approved):

PR #409, opened 2022-10-27
Solved homework

CyberDogFK:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,85 @@
...
+    private Object createNewInstance(Class<?> clazz) {
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        Object instance;
+        try {
+            instance = clazz.getConstructor().newInstance();
+            instances.put(clazz, instance);
+        } catch (InstantiationException | IllegalAccessException
+                 | InvocationTargetException | NoSuchMethodException e) {
```
nazariipertsak, line 65: You can change 
```suggestion
        } catch (ReflectiveOperationException e) {
```
That was in common mistakes

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,85 @@
...
+    private final Map<Class<?>, Object> instances = new HashMap<>();
 
     public static Injector getInjector() {
+        implementationMap = fillDefaultStorage();
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        if (!interfaceClazz.isInterface()
```
nazariipertsak, line 25: Do you need this check? You will not be able to create object of Interface in any way, only of it's implementation. So interface never needs annotation @Component

nazariipertsak (changes requested):

nazariipertsak (approved):

src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
ChumakAn, line 6: You also should practice with Target annotation

ChumakAn (changes requested):

ChumakAn (approved):

PR #411, opened 2022-10-28
Injector implemented

hpavlo:


YuriiHryniv (approved):

src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
liliia-ponomarenko, line 6: You also should practice with Target annotation

liliia-ponomarenko (approved):
Good job! 🏆😎

PR #413, opened 2022-10-29
implement an injector that uses @Component and @Inject annotations

Bodddd:


shmex-j (approved):
LGTM!

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,7 @@
 package mate.academy.lib;
 
-public @interface Inject {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
-}
+@Retention(RetentionPolicy.RUNTIME)
```
ChumakAn, line 6: it's good to use target here too, as it can be applied to fields

ChumakAn (changes requested):
Good! Minor NIT

ChumakAn (approved):

PR #415, opened 2022-10-31
implemented task

vov4ik89:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Component {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Component {
```
arthur4655, line 7: I think you can add Retention annotation here

src/main/java/mate.academy/service/ProductService.java
```diff
@@ -1,8 +1,10 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
```
arthur4655, line 7: I'm not sure that this is neсessary  to add Component annotation here

src/main/java/mate.academy/service/ProductParser.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
```
arthur4655, line 6: and here

src/main/java/mate.academy/service/FileReaderService.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 
+@Component
```
arthur4655, line 6: and here

arthur4655 (changes requested):

src/main/java/mate.academy/service/ProductService.java
vov4ik89: when I pick up the annotation, the tests do not pass

arthur4655 (approved):
LGTM

PR #423, opened 2022-11-06
Realized Hm with injector

OleksandrKorniychuk:


igorgalenko (approved):
Good job!

src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Component {
```
hronograf, line 7: we can also use `Target` annotation

hronograf (approved):
LGTM! 🔥

PR #433, opened 2022-11-08
task completed, injector implemented, @Component, @Inject added

yehbreus:


src/main/java/mate.academy/Main.java
```diff
@@ -10,7 +10,7 @@ public class Main {
     public static void main(String[] args) {
         // Please test your Injector here. Feel free to push this class as a part of your solution
```
GHSpeedWagon, line 11: redundant comment

delete it in other casses

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
     private static final Injector injector = new Injector();
+    private Map<Class<?>, Object> instances = new HashMap<>();
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementationInstance = null;
```
GHSpeedWagon, line 23: can we not initialize this field 

(dont use null)

GHSpeedWagon (changes requested):

src/main/java/mate.academy/lib/Injector.java
yehbreus: no, we must initialize it, because if we dont do it - this object will be out of scope

GHSpeedWagon (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+        Class<?> clazz = findImplementation(interfaceClazz);
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Injection failed, missing @Component "
+                    + "annotation on the class " + clazz.getName());
+        }
+        Field[] declaredFields = interfaceClazz.getDeclaredFields();
+        for (Field field : declaredFields) {
+            if (field.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = getInstance(field.getType());
+                clazzImplementationInstance = createNewInstance(clazz);
```
hronograf, line 33: can we initialize `clazzImplementationInstance` before `for` loop?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+                    field.set(clazzImplementationInstance, fieldInstance);
+                } catch (IllegalAccessException e) {
+                    throw new RuntimeException("Can't initialize field value. Class: "
+                            + clazz.getName() + ". Field: " + field.getName());
+                }
+            }
+        }
+        if (clazzImplementationInstance == null) {
+            clazzImplementationInstance = createNewInstance(clazz);
+        }
```
hronograf, line 45: so we will not need this `if`

src/main/java/mate.academy/lib/Injector.java
hronograf, line 23: and this `= null`

hronograf (changes requested):
Good!

hronograf (approved):
Nice! 💎

PR #434, opened 2022-11-08
My injector impl

anton-pushkar:


src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
```
GHSpeedWagon, line 5: try to use @Target

GHSpeedWagon (approved):
check my advise

liliia-ponomarenko (approved):
LGTM! 🔥

PR #435, opened 2022-11-08
Injector implemented.

luka-lozovyk:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,70 @@
...
+            throw new RuntimeException("Can't create an instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = Map.of(
+                FileReaderService.class, FileReaderServiceImpl.class,
+                ProductParser.class, ProductParserImpl.class,
+                ProductService.class, ProductServiceImpl.class);
+        return interfaceImplementations.get(interfaceClazz);
```
Mar4uk21, line 68: Imagine if the class (not interface) comes to the entrance, I think we need a check here for this situation

Mar4uk21 (changes requested):

Mar4uk21 (approved):
LGTM)

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+                        field.set(clazzImplInstance, fieldInstance);
+                    } catch (IllegalAccessException e) {
+                        throw new RuntimeException("Can't initialize field value. "
+                                + "Class: " + clazz.getName() + ". Field: " + field.getName());
+                    }
+                }
+            }
+        } else {
+            throw new RuntimeException("Class " + clazz + " does not have @Component annotation!");
+        }
```
hronograf, line 42: in cases when we should check something and throw an exception or return a value
it is a good practice to do it as early as possible
read [this article](https://medium.com/swlh/return-early-pattern-3d18a41bba8) to learn more

hronograf (changes requested):
Good!

hronograf (approved):
Good! 👍

PR #438, opened 2022-11-08
hw-dependency-injection

Iryna-a:


src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
NSay0, line 6: We can also mark @Target(ElementType=...) to determine kinds of elements an annotation interface can be applied to.

NSay0 (approved):
LGTM!

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+            if (field.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = getInstance(field.getType());
+                clazzImplementationInstance = createNewInstance(clazz);
+                try {
+                    field.setAccessible(true);
+                    field.set(clazzImplementationInstance, fieldInstance);
+                } catch (IllegalAccessException e) {
+                    throw new RuntimeException("Can't initiate field value. Class: "
+                                + clazz.getName() + ". Field: " + field.getName());
+                }
```
hronograf, line 40: image

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
+        }
```
hronograf, line 71: same

hronograf (changes requested):

hronograf (approved):
Great! 💥

PR #440, opened 2022-11-08
released injector for productService

apc10bbe:


src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
```
GHSpeedWagon, line 5: try to use @Target

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Can't get instance. Class: "
+                    + clazz + " does not have annotation @Component.");
+        }
+        Field[] declaredFields = clazz.getDeclaredFields();
+        Object clazzImplementationInstance = null;
+        for (Field field : declaredFields) {
+            if (field.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = getInstance(field.getType());
+                clazzImplementationInstance = createNewinstance(clazz);
```
GHSpeedWagon, line 33: we can firstly chack do we have @Component in class that we are initialize

if class that we will initialize have @Component 
Object fieldInInstance
clazz...
for...

GHSpeedWagon (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+                try {
+                    field.setAccessible(true);
+                    field.set(clazzImplementationInstance, fieldInstance);
+                } catch (IllegalAccessException e) {
+                    throw new RuntimeException("Can't initialize field value. " + "Class: "
+                            + clazz.getName() + ". Field: " + field.getName());
```
liliia-ponomarenko, line 39:
```suggestion
                            + clazz.getName() + ". Field: " + field.getName(), e);
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
```
liliia-ponomarenko, line 59:
```suggestion
            throw new RuntimeException("Can't create a new instance of " + clazz.getName(), e);
```

liliia-ponomarenko (approved):
great 😊  but don't forget to throw an exception

PR #443, opened 2022-11-09
Implemented DI

sequencerr:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,69 @@
...
-        return null;
+        Class<?> clazz = getImplementation(interfaceClazz);
+        Arrays.stream(clazz.getDeclaredFields())
+                .filter(field -> field.isAnnotationPresent(Inject.class))
+                .forEach(field -> {
+                    field.setAccessible(true);
+                    try {
+                        field.set(
+                                createNewInstance(clazz),
+                                getInstance(field.getType())
```
GHSpeedWagon, line 36: where do we use @Component annotation?

GHSpeedWagon (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Class<?> clazz = getImplementation(interfaceClazz);
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException(interfaceClazz.getName()
+                    + "Is not supported. (Must annotated with Component)");
+        }
```
GHSpeedWagon, line 32: gj

GHSpeedWagon (approved):
good job

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+                .filter(field -> field.isAnnotationPresent(Inject.class))
+                .forEach(field -> {
+                    field.setAccessible(true);
+                    try {
+                        field.set(
+                                createNewInstance(clazz),
+                                getInstance(field.getType())
+                        );
+                    } catch (IllegalAccessException e) {
+                        throw new RuntimeException("Error while injecting instance into field", e);
```
liliia-ponomarenko, line 43: Would  be better to give pass parameters in the exception message

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+    private static Object createNewInstance(Class<?> clazz) {
+        Object foundInstance = instancedImplementations.get(clazz);
+        if (foundInstance != null) {
+            return foundInstance;
+        }
+        try {
+            Object instance = clazz.getConstructor().newInstance();
+            instancedImplementations.put(clazz, instance);
+            return instance;
+        } catch (Exception e) {
```
liliia-ponomarenko, line 58:
```suggestion
        } catch (ReflectiveOperationException e) {
```
The same about parameters. Also, use ReflectiveOperationException
image

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+
+    private static Class<?> getImplementation(Class<?> interfaceClazz) {
+        return interfaceClazz.isInterface()
+               ? interfaceImplementations.get(interfaceClazz)
+               : interfaceImplementations
+                       .keySet()
+                       .stream()
+                       .filter(interfaceClazz::equals)
+                       .findFirst()
+                       .orElseThrow();
```
liliia-ponomarenko, line 71: Why do you need this logic? you should  return the input class if ```interfaceImplementations``` haven’t a corresponding  implementation

liliia-ponomarenko (changes requested):
good job! Let’s  improve your solution ;)

src/main/java/mate.academy/lib/Injector.java
sequencerr: It breaks test cases, because no, I should return only for supported classes

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private static final Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
+    private static final Map<Class<?>, Object> instancedImplementations = new HashMap<>();
 
     public static Injector getInjector() {
```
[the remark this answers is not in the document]
sequencerr: Because we need to get instance of singleton outside the class

liliia-ponomarenko (approved):
good job! 👍

PR #445, opened 2022-11-09
implemented injector that uses "Component" and "Inject"

Yaroslav781:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
```
anastasiavasylyk, line 5: You may add @Target, but it optional

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can`t create a new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementation = new HashMap<>();
```
anastasiavasylyk, line 65: Use Map.of()

anastasiavasylyk (changes requested):
little fixes

anastasiavasylyk (approved):
Good!

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+            if (field.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = getInstance(field.getType());
+                clazzImplementationInstance = createNewInstance(clazz);
+                field.setAccessible(true);
+                try {
+                    field.set(clazzImplementationInstance, fieldInstance);
+                } catch (IllegalAccessException e) {
+                    throw new RuntimeException("Can`t initialize field value. "
+                            + "Class: " + clazz.getName()
+                            + ". Field: " + field.getName());
```
liliia-ponomarenko, line 40:
```suggestion
                            + ". Field: " + field.getName(), e);
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can`t create a new instance of " + clazz.getName());
```
liliia-ponomarenko, line 60:
```suggestion
            throw new RuntimeException("Can`t create a new instance of " + clazz.getName(), e);
```

liliia-ponomarenko (approved):
great 😊 but don't forget to throw an exception

PR #451, opened 2022-11-21
Solution was added.

KatyaHumeniuk:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+            }
+        }
+        throw new RuntimeException("Can't create a new instance of " + clazz.getName());
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
```
aaananas, line 67: to optimize your code you can make this map global
an initialize it with fabric method Maps.of(key, value, key, value...)
to avoid creating and initializing this map each time this method is invoked

Maps.of(..) method creates immutable map - that cant be changed in future

aaananas (approved):

PR #453, opened 2022-12-09
Solution

kateryna-mykh:


aleksandr-bevza (changes requested):
I cloned this project to my laptop and run. Tests are green but "main" method doesn't work.
Your annotation "Inject" don't work correctly. It doesn't injects fields because of wrong implementation of "getInstance" method. This should be reworked.

Please make your method "main" also work

zachar-chorny (approved):

PR #459, opened 2022-12-19
method getInstance in the class Injector was apdated

DanyloBorysenko:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Component {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Component {
```
aleksandr-bevza, line 7: Add Target

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
```
aleksandr-bevza, line 67: lets make this a class-level variable and create just 1 map. because this map is created each time this method is invoked
you can use Maps.of(key, value, key, value ...) - this method creates an immutable map

private final Map<Class<?>, Class<?>> interfaceImplementations = Maps.of(

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+        Field[] declaredFields = clazz.getDeclaredFields();
+        for (Field field : declaredFields) {
+            if (field.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = getInstance(field.getType());
+                clazzImplementationInstance = createNewInstance(clazz);
+                field.setAccessible(true);
+                try {
+                    field.set(clazzImplementationInstance, fieldInstance);
+                } catch (IllegalAccessException e) {
+                    throw new RuntimeException("Can't initialize field value. Class "
```
aleksandr-bevza, line 38: Include field name to this message

aleksandr-bevza (changes requested):

aleksandr-bevza (approved):

PR #470, opened 2023-01-04
implemented methods in Injector class

metelska-anastasiia:
put annotations where needed
Injector was checked in main method

### DIFF
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,13 +1,73 @@
 package mate.academy.lib;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.util.HashMap;
+import java.util.Map;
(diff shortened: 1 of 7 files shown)


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        Field[] declaredFields = clazz.getDeclaredFields();
```
zachar-chorny, line 25: What if annotation "Component" is not present in class?

zachar-chorny (changes requested):

zachar-chorny (approved):

PR #474, opened 2023-01-05
own injector was realized

Roman-Usyk:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,79 @@
...
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException(CREATE_NEW_INSTANCE_ERROR + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementation = new HashMap<>();
+        interfaceImplementation.put(ProductService.class, ProductServiceImpl.class);
+        interfaceImplementation.put(FileReaderService.class, FileReaderServiceImpl.class);
```
aaananas, line 72: you can make this map a class-level variable

also you can use Map.of(key, value, key, value) method - it will create a immutable collection

aaananas (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can`t create new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        interfaceImplementation.put(ProductService.class, ProductServiceImpl.class);
+        interfaceImplementation.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementation.put(ProductParser.class, ProductParserImpl.class);
+        if (interfaceClazz.isInterface()) {
```
aaananas, line 68:
```suggestion
        if (interfaceClazz.isInterface()) {
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private final Map<Class<?>, Object> instances = new HashMap<>();
+    private final Map<Class<?>, Class<?>> interfaceImplementation = new HashMap<>();
```
```suggestion
    private final Map<Class<?>, Class<?>> interfaceImplementation = Maps.of(
                               ProductService.class, ProductServiceImpl.class, 
                               FileReaderService.class, FileReaderServiceImpl.class, 
                               ProductParser.class, ProductParserImpl.class);
```

aaananas (approved):

PR #476, opened 2023-01-07
Injector was implemented

katrevichdv:
Injector was implemented.
source code package /mate.academy was moved to mate/academy/


src/main/java/mate/academy/Main.java
```diff
@@ -8,9 +8,8 @@
 public class Main {
 
     public static void main(String[] args) {
```
karma-o, line 10: I can see that you have changed the structure of the project
`mate.academy` were 2 packages, but now they are directories `mate/academy`, so let's change it back to how it was.

karma-o (changes requested):

src/main/java/mate/academy/Main.java
katrevichdv: > I can see that you have changed the structure of the project `mate.academy` were 2 packages, but now they are directories `mate/academy`, so let's change it back to how it was.
You propably should revise java specification

kozhukhovsky (approved):

PR #477, opened 2023-01-09
Dependency Injection solution

diana-lomei:


src/main/java/mate.academy/Main.java
```diff
@@ -10,7 +10,7 @@ public class Main {
     public static void main(String[] args) {
         // Please test your Injector here. Feel free to push this class as a part of your solution
         Injector injector = Injector.getInjector();
-        ProductService productService = null;
+        ProductService productService = (ProductService) injector.getInstance(ProductService.class);
```
aleksandr-bevza, line 13: Injector not working. If you run method "main" - you will see that injection didn't happen. Fields of ProductService were not injected.
Fix a mistake in your "Injector.class". You should take declared fields from class but not from interface

aleksandr-bevza (changes requested):

nacenik (approved):
Good job!

PR #479, opened 2023-01-13
done

Richaa2:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
okuzan, line 6: add @Target

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
okuzan, line 6: add @Target

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,71 @@
...
+                        + clazz.getName());
+            }
+            if (field.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = getInstance(field.getType());
+                clazzImplementationInstance = createNewInstance(clazz);
+                try {
+                    field.setAccessible(true);
+                    field.set(clazzImplementationInstance, fieldInstance);
+                } catch (IllegalAccessException e) {
+                    throw new RuntimeException("Can't initialize field value. Class: "
```
okuzan, line 38: include root exception

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,71 @@
...
+            instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create instance: " + clazz.getName() + e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
```
okuzan, line 65: make this map a class-level constant to avoid unnecessary recreation per each method invocation, you can use Map.of()

okuzan (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,70 @@
...
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private final Map<Class<?>, Object> instances = new HashMap<>();
+    private final Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
```
okuzan, line 17:
```suggestion
    private final Map<Class<?>, Class<?>> interfaceImplementations = Map.of(...;
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,70 @@
...
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create instance: " + clazz.getName() + e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
```
okuzan, line 67: initialize hashmap with these key-values in the declaration - you need to do it only once, now you do it every time this method is called

Richaa2: thanks now i understand

okuzan (approved):
Good job!

PR #483, opened 2023-01-16
created Injector

InnaDiachkina:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,72 @@
...
+        if (interfaceClazz.isInterface()) {
+            return interfaceImplementations.get(interfaceClazz);
+        }
+        return interfaceClazz;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Can't create new instance of " + clazz.getName());
+        }
```
Maks4u, line 59: We can do it at the start of getInstance() method (after using clazz implementation). It's prevent us to do unnecessary steps (like call getInstance for other objects, trigger create newInstance, go to loop etc.)

Maks4u (approved):

Nick97-git (approved):

PR #488, opened 2023-01-17
implementation created

solomiia-voloshyn:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Component {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Component {
```
yurii-dikalo, line 7: you can add Target meta-annotation to specify that this Component should only be used with classes, in this case, its value should be ElementType.Type

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Inject {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Inject {
```
yurii-dikalo, line 7: same here, you can use Target. I don't know if you have to, though, but this makes it more specified

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+            throw new RuntimeException("Can`t create a new instance of "
+                    + clazz.getName(),e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementation = new HashMap<>();
+        interfaceImplementation.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementation.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementation.put(ProductService.class, ProductServiceImpl.class);
```
yurii-dikalo, line 68: you don't need to create this map each time you invoke this method (because you do it a lot), so you can initialize it when it is declared as a field using map.of

yurii-dikalo (changes requested):

src/main/java/mate.academy/lib/Injector.java
solomiia-voloshyn: ✅

src/main/java/mate.academy/lib/Inject.java
solomiia-voloshyn: ✅

src/main/java/mate.academy/lib/Component.java
solomiia-voloshyn: ✅

yurii-dikalo (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private final Map<Class<?>, Class<?>> interfaceImplementations = Map.of(
+            FileReaderService.class, FileReaderServiceImpl.class,
+            ProductParser.class, ProductParserImpl.class,
+            ProductService.class, ProductServiceImpl.class
+    );
```
Nick97-git, line 20: let's make it static and initialize in a static block of initialization

Nick97-git (changes requested):

Nick97-git (approved):

PR #491, opened 2023-01-17
Injector implementation 1 commit

ar4ibaldd:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+            return instances.get(clazz);
+        }
+
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                 | InstantiationException | IllegalAccessException e) {
```
MaxMaksym, line 61:
```suggestion
        } catch (ReflectiveOperationException e) {
```

MaxMaksym (approved):

src/main/java/mate.academy/lib/Injector.java
Nick97-git: make as suggested

Nick97-git (changes requested):
I don't see changed Main class where you should use Injector

Nick97-git (approved):

PR #492, opened 2023-01-18
tasks were implemented

lysytskapolina:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,78 @@
...
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
-    private static final Injector injector = new Injector();
+    private static final Injector INJECTOR = new Injector();
```
taras-zhyravetskyi, line 15: I think it's not a constant, so field name should be in lowercase

taras-zhyravetskyi (approved):

src/main/java/mate.academy/lib/Injector.java
Nick97-git: agree

Nick97-git (approved):

PR #502, opened 2023-01-22
injector ready

ItsUnbidden:


pom.xml
```diff
@@ -38,14 +38,6 @@
...
                 <artifactId>maven-checkstyle-plugin</artifactId>
                 <version>3.1.1</version>
-                <executions>
-                    <execution>
-                        <phase>compile</phase>
-                        <goals>
-                            <goal>check</goal>
-                        </goals>
-                    </execution>
-                </executions>
```
Rommelua, line 48: do not change pom.xml

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "java.configuration.updateBuildConfiguration": "interactive"
+}
```
Rommelua, line 3: what is it for?

Rommelua (changes requested):
Do not rename package
image

pom.xml
ItsUnbidden: I tried fixing this, but my program doesn't want to compile. It appears that it thinks 'mate.academy' is not one folder, but two and so it can't find the package. pom.xml causes problems with the tests and they don't want to run for some reason.

.vscode/settings.json
ItsUnbidden: This got pushed here accidentally. It has nothing to do with the program.

aaananas (approved):

PR #504, opened 2023-01-23
Here is my solution for hw-dependency-injection

andmalenko:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                 | InstantiationException | IllegalAccessException e) {
```
aaananas, line 56: those exceptions extend the same exception, lets catch that exception

aaananas (changes requested):

src/main/java/mate.academy/lib/Injector.java
aaananas: not fixed

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private Map<Class<?>, Object> instances = new HashMap<>();
+    private Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
```
aaananas, line 18:
```suggestion
    private Map<Class<?>, Class<?>> interfaceImplementations = Map.of(
                                FileReaderService.class, FileReaderServiceImpl.class,
                                ProductParser.class, ProductParserImpl.class,
                                ProductService.class, ProductServiceImpl.class);
```

aaananas (changes requested):

src/main/java/mate.academy/lib/Injector.java
andmalenko: now understand what you mean :)

src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
Rommelua, line 6: adding Target is a good practice

Rommelua (approved):

PR #508, opened 2023-01-27
task done

crowlynx:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+                }
+            }
+        }
+        if (clazzImplementationInstance == null) {
+            clazzImplementationInstance = createNewInstance(clazz);
+        }
+        return clazzImplementationInstance;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
```
karma-o, line 54:
```suggestion
    private Object getOrCreateNewInstance(Class<?> clazz) {
```

src/main/java/mate.academy/Main.java
```diff
@@ -4,13 +4,14 @@
 import mate.academy.lib.Injector;
 import mate.academy.model.Product;
 import mate.academy.service.ProductService;
+import mate.academy.service.impl.ProductServiceImpl;
 
 public class Main {
 
     public static void main(String[] args) {
-        // Please test your Injector here. Feel free to push this class as a part of your solution
         Injector injector = Injector.getInjector();
```
karma-o, line 12: can be moved to a class-level field, but it is up to you

karma-o (changes requested):

src/main/java/mate.academy/Main.java
```diff
@@ -7,9 +7,10 @@
 import mate.academy.service.impl.ProductServiceImpl;
 
 public class Main {
+    private static Injector injector;
```
karma-o:
```suggestion
    private Injector injector = Injector.getInjector();
```

karma-o (approved):

PR #520, opened 2023-02-10
Dependency injection solution

nkharyshyn:
Wrote logic for injector, added Component annotation to implementations, instances of which should be created by injector


Ivan95kos (approved):
Great job

PR #523, opened 2023-02-15
Initial commit

waspp01:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,79 @@
...
+            throw new RuntimeException("Can't create a new instance of "
+                    + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
```
aaananas, line 73: lets make this map a class-level variable to avoid creating this map each time this method is invoked
lets create it and initialize it using a meyhod Maps.of(key, value, key, value...) - this creates an immutable map

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,79 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                | InstantiationException | IllegalAccessException e) {
```
aaananas, line 63: see common mistakes to fix it

aaananas (changes requested):

Rommelua (approved):

PR #536, opened 2023-03-03
Added solution for jv-dependency-injection

serhii-shyian:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,90 @@
...
+    private static Injector injector;
+    private static final String ANNOTATION = "Component";
+
+    public Injector() {
+        implementations = Map.of(
+                FileReaderService.class, FileReaderServiceImpl.class,
+                ProductParser.class, ProductParserImpl.class,
+                ProductService.class, ProductServiceImpl.class);
+        instances = new HashMap<>();
+    }
```
okuzan, line 26: no need for the constructor, move Map.of to declaration

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,90 @@
...
+        if (interfaceClazz.isInterface()) {
+            Class<?> implementationClazz = implementations.get(interfaceClazz);
+            if (!implementationClazz.isAnnotationPresent(Component.class)) {
+                throw new RuntimeException("Can't find annotation "
+                                + ANNOTATION + " for class "
+                                + implementationClazz.getName() + ".");
+            }
+            return implementationClazz;
+        }
+        return interfaceClazz;
```
okuzan, line 88:
```suggestion
    if (interfaceClazz.isInterface()) {
        return implementations.get(interfaceClazz);
    }
    return interfaceClazz;
```

okuzan (changes requested):

kozhukhovsky (approved):

PR #538, opened 2023-03-08
hw-solutions

nick0226:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+                    + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
+        if (interfaceClazz.isInterface()) {
```
aaananas, line 68: lets move this map to a class level variable and initialize it there - to avoid creating this map each time this method is invoked

aaananas (changes requested):

andriy-shulepa (approved):

PR #539, opened 2023-03-09
I implement Injector in this solution. I use Injector with @Inject an…

NazarZva:
…d @Component annotations.

### DIFF
--- a/src/main/java/mate.academy/lib/Inject.java
+++ b/src/main/java/mate.academy/lib/Inject.java
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Inject {
 
 }
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,13 +1,73 @@
 package mate.academy.lib;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.util.HashMap;
+import java.util.Map;
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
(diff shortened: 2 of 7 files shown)


src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
nacenik, line 6: Can you use this annotation for classes? Or only for field? [See here ](https://docs.oracle.com/javase/8/docs/api/java/lang/annotation/Target.html)
add target to all custom annotations

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("can`t create new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementation = new HashMap<>();
+        interfaceImplementation.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementation.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementation.put(ProductService.class, ProductServiceImpl.class);
```
nacenik, line 67: do you really need to create this map for every single check? I'd recommend make this map a class-level variable - you can initialize it with Map.of() method

nacenik (changes requested):

src/main/java/mate.academy/lib/Inject.java
nacenik: it's a good practice to add `Target` for all custom annotations. `Inject` is not an exclusion :)

nacenik (approved):
LGTM! 🚀

PR #542, opened 2023-03-13
hw dependency injection have done

ArtemMakovskyy:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+            clazzImplementationInstance = createNewInstance(clazz);
+        }
+        return clazzImplementationInstance;
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementation = new HashMap<>();
+        interfaceImplementation.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementation.put(ProductService.class, ProductServiceImpl.class);
+        interfaceImplementation.put(ProductParser.class, ProductParserImpl.class);
```
Rommelua, line 54: Make this map a class field. Fill it in once (not inside this method).

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
-    private static final Injector injector = new Injector();
+    private static final Injector INJECTOR = new Injector();
+    private static final Map<Class<?>, Object> INSTANCES = new HashMap<>();
```
Rommelua, line 16: camelCase. These are not constants.

Rommelua (changes requested):

src/main/java/mate.academy/lib/Injector.java
ArtemMakovskyy: Bogdan in the video uses "private static final Injector injector = new Injector();" I'm confused. It looks like a constant and should be in the upper case. Now I see that "static final" is not necessarily a constant. Leonid, can you tell me what the rule is here?

Rommelua: https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names
If we can change inner state of the object - it is NOT constant
image

Rommelua (approved):

PR #545, opened 2023-03-16
Solution for DI

oleksandrCod:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
okuzan, line 6: add target here and in Inject

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,80 @@
...
+                 | IllegalAccessException e) {
+            throw new RuntimeException("Can't create new  instance of: " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImpl(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImpl = new HashMap<>();
+        interfaceImpl.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImpl.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImpl.put(ProductService.class, ProductServiceImpl.class);
```
okuzan, line 74: make this class constant. initialize using Map.of for example

okuzan (changes requested):

olekskov (approved):
Nice job!

PR #548, opened 2023-03-17
Copy-Paste HomeWork Success!

IlkoGood:


yurihhhT (approved):
Good joob)
Good name of pull as well)

src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Component {
```
nacenik, line 7: Do not forget to add @Target to custom annotations.

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Inject {
```
nacenik, line 7: Do not forget to add @Target to custom annotations.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,72 @@
...
+                 | IllegalAccessException | InvocationTargetException e) {
+            throw new RuntimeException("Can't create new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> implMap = new HashMap<>();
+        implMap.put(FileReaderService.class, FileReaderServiceImpl.class);
+        implMap.put(ProductParser.class, ProductParserImpl.class);
+        implMap.put(ProductService.class, ProductServiceImpl.class);
```
nacenik, line 69: Make Interface Implementations map a class field. You can fill it in using Map.of().

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,72 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InstantiationException
+                 | IllegalAccessException | InvocationTargetException e) {
```
nacenik, line 60: You can put just `ReflectiveOperationException` instead of all exceptions that you wrote down. [See here](https://docs.oracle.com/javase/7/docs/api/java/lang/ReflectiveOperationException.html)

nacenik (changes requested):

src/main/java/mate.academy/lib/Component.java
IlkoGood: so I need to annotate that my annotation belongs to annotation_type, OK)

nacenik (approved):
LGTM! 🚀

PR #550, opened 2023-03-18
Hw-injection

KosMarch:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create new instance of " + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementation = new HashMap<>();
+        interfaceImplementation.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementation.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementation.put(ProductService.class, ProductServiceImpl.class);
```
denys-mg, line 63: Make it a class field and initialize using Map.of() or by another way (but not using a constructor). The problem is you recreate and fill 'interfaceImplementation' each time when you invoke 'findImplementation' method

denys-mg (changes requested):
Good job, left you my remarks)

denys-mg (approved):

src/main/java/mate.academy/Main.java
```diff
@@ -1,17 +1,7 @@
...
-import mate.academy.service.ProductService;
-
 public class Main {
 
     public static void main(String[] args) {
-        // Please test your Injector here. Feel free to push this class as a part of your solution
-        Injector injector = Injector.getInjector();
-        ProductService productService = null;
-        List<Product> products = productService.getAllFromFile("products.txt");
-        products.forEach(System.out::println);
```
nacenik, line 15: Why did you delete all logic from here? You need to change only one line.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
 public class Injector {
     private static final Injector injector = new Injector();
+    private static final Map<Class<?>, Object> instances = new HashMap<>();
+    private static final Map<Class<?>, Class<?>> interfaceImplementation = new HashMap<>();
+
+    static {
+        interfaceImplementation.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementation.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementation.put(ProductService.class, ProductServiceImpl.class);
+    }
```
nacenik, line 23: Fill the map inline. You can fill it in using Map.of().

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private static final Map<Class<?>, Object> instances = new HashMap<>();
+    private static final Map<Class<?>, Class<?>> interfaceImplementation = new HashMap<>();
```
nacenik, line 17: It's a good idea to make this fields private, final, and non-static

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create new instance of " + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (!interfaceImplementation.get(interfaceClazz).isAnnotationPresent(Component.class)) {
```
nacenik, line 67: what happens if `get()` returns `null`?

nacenik (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,80 @@
...
 public class Injector {
     private static final Injector injector = new Injector();
+    private final Map<Class<?>, Object> instances = new HashMap<>();
+    private final Map<Class<?>, Class<?>> interfaceImplementation;
+
+    {
+        interfaceImplementation = Map.of(FileReaderService.class, FileReaderServiceImpl.class,
+                ProductParser.class, ProductParserImpl.class,
+                ProductService.class, ProductServiceImpl.class);
+    }
```
nacenik, line 23: Do you really need this block? You can do the same in 17 line without this block

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,80 @@
...
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create new instance of " + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (interfaceImplementation.get(interfaceClazz) == null) {
```
nacenik, line 67: not exactly, you can have another class, that can contain @Component annotation. 
Check if your map contains a key, if yes - just return the value, if no - create new one.

src/main/java/mate.academy/lib/Injector.java
nacenik: Not exactly, see other comment

nacenik (changes requested):
Fix some logic issues

sarakhmen (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create new instance of " + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (!interfaceImplementation.containsKey(interfaceClazz)) {
+            throw new RuntimeException("Can't find implementation of this interface "
+                    + interfaceClazz);
+        }
```
KseniiaMakarova, line 70: if this is a class, not an interface, will map contain the key? is it ok to throw exception in this case?

KseniiaMakarova (approved):

PR #552, opened 2023-03-18
solved task

KyslyiViacheslav:
implemented an injector that uses @Component and @Inject

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,6 +26,11 @@
             <scope>test</scope>
             <version>5.5.2</version>
         </dependency>
+        <dependency>
+            <groupId>org.reflections</groupId>
+            <artifactId>reflections</artifactId>
+            <version>0.10.2</version>
+        </dependency>
     </dependencies>
     <build>
         <plugins>
--- a/src/main/java/mate.academy/lib/Inject.java
+++ b/src/main/java/mate.academy/lib/Inject.java
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Inject {
 
 }
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,13 +1,91 @@
 package mate.academy.lib;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+import org.reflections.Reflections;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private final Map<Class<?>, Object> instances = new HashMap<>();
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Class<?> clazz = findImplV2(interfaceClazz);
+        isImplClassWithAnotation(clazz);
+        Object clazzImplInstance = null;
+        Field[] clazzDeclaratedFields = clazz.getDeclaredFields();
+        for (Field clazzField : clazzDeclaratedFields) {
+            if (clazzField.isAnnotationPresent(Inject.class)) {
+                Object clazzFieldInstance = getInstance(clazzField.getType());
+                clazzImplInstance = createNewInstance(clazz);
+                clazzField.setAccessible(true);
+                try {
+                    clazzField.set(clazzImplInstance,clazzFieldInstance);
+                } catch (IllegalAccessException e) {
+                    throw new RuntimeException("Can't initialize field value"
+                            + " Class: " + clazz.getName()
+                            + " Field: " + clazzField.getName(),e);
+                }
+            }
+        }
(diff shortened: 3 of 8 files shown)


src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
```
AndrewkeY-hub, line 5: I thik it would be better to add annotation @Target(ElementType.FIELD)

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,91 @@
...
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create instance of : " + clazz.getName(),e);
+        }
+    }
+
+    private Class<?> findImpl(Class<?> clazz) {
+        Map<Class<?>,Class<?>> implClazzez = new HashMap<>();
```
AndrewkeY-hub, line 73: Separate like this: Map<Class<?>, Class<?>>

AndrewkeY-hub (approved):

pom.xml
```diff
@@ -26,6 +26,11 @@
             <scope>test</scope>
             <version>5.5.2</version>
         </dependency>
+        <dependency>
+            <groupId>org.reflections</groupId>
+            <artifactId>reflections</artifactId>
+            <version>0.10.2</version>
+        </dependency>
```
nacenik, line 33: You don't need to add any type of additional dependency

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,91 @@
...
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create instance of : " + clazz.getName(),e);
+        }
+    }
+
+    private Class<?> findImpl(Class<?> clazz) {
+        Map<Class<?>, Class<?>> implClazzez = new HashMap<>();
+        implClazzez.put(ProductService.class, ProductServiceImpl.class);
+        implClazzez.put(ProductParser.class, ProductParserImpl.class);
+        implClazzez.put(FileReaderService.class, FileReaderServiceImpl.class);
```
nacenik, line 76: Make Interface Implementations map a class field. You can fill it in using Map.of().

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,91 @@
...
+        return clazz;
+    }
+
+    private Class<?> findImplV2(Class<?> clazz) {
+        Reflections reflections = new Reflections("mate.academy");
+        Set<Class<?>> classes = (Set<Class<?>>) reflections.getSubTypesOf(clazz);
+        if (clazz.isInterface()) {
+            return classes.iterator().next();
+        }
+        return clazz;
```
nacenik, line 89: don't use the additional library.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,91 @@
...
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create instance of : " + clazz.getName(),e);
+        }
+    }
+
+    private Class<?> findImpl(Class<?> clazz) {
```
nacenik, line 72: don't use short for in method names

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,91 @@
...
+                }
+            }
+        }
+        if (clazzImplInstance == null) {
+            clazzImplInstance = createNewInstance(clazz);
+        }
+        return clazzImplInstance;
+    }
+
+    private boolean isImplClassWithAnotation(Class<?> clazz) {
```
nacenik, line 49: Do you really need this method? If you think yes - do you really need to return the type as a boolean? 
Don't use the short forms of words in method names.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,91 @@
...
-        return null;
+        Class<?> clazz = findImplV2(interfaceClazz);
+        isImplClassWithAnotation(clazz);
+        Object clazzImplInstance = null;
+        Field[] clazzDeclaratedFields = clazz.getDeclaredFields();
+        for (Field clazzField : clazzDeclaratedFields) {
+            if (clazzField.isAnnotationPresent(Inject.class)) {
+                Object clazzFieldInstance = getInstance(clazzField.getType());
+                clazzImplInstance = createNewInstance(clazz);
+                clazzField.setAccessible(true);
```
nacenik, line 33: put this line inside the try block, cause `setAccessible()` would throw an exception too. [See documentation](https://docs.oracle.com/javase/7/docs/api/java/lang/reflect/AccessibleObject.html#setAccessible(boolean))

nacenik (changes requested):

src/main/java/mate.academy/lib/Injector.java
KyslyiViacheslav: ++

src/main/java/mate.academy/lib/Injector.java
KyslyiViacheslav: ++

src/main/java/mate.academy/lib/Injector.java
KyslyiViacheslav: ++

src/main/java/mate.academy/lib/Injector.java
KyslyiViacheslav: ++

PR #555, opened 2023-03-19
Solution of Dependency Injection

AnonimToster:


AlexandLes (approved):
Good solution!

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,89 @@
...
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
+    private static final Map<Class<?>, Class<?>> implementations = new HashMap<>();
+    private static final Map<Class<?>, Object> components = new HashMap<>();
```
nacenik, line 17: this map shouldn't be static, just private and final

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,89 @@
...
+    }
+
+    private static void fillMap() {
+        implementations.put(FileReaderService.class,
+                FileReaderServiceImpl.class);
+        implementations.put(ProductParser.class,
+                ProductParserImpl.class);
+        implementations.put(ProductService.class,
+                ProductServiceImpl.class);
+    }
```
nacenik, line 31: You don't need to create an additional method. You can fill it in using `Map.of().`

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,89 @@
...
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Class<?> implementation = getImpl(interfaceClazz);
+        Object clazzImplementationsInstance = instanceOf(implementation);
+        Field[] fields = implementation.getDeclaredFields();
+        for (Field field : fields) {
+            if (field.isAnnotationPresent(Inject.class)) {
+                Object value = getInstance(field.getType());
+                field.setAccessible(true);
```
nacenik, line 44: put this line inside the try block, cause `setAccessible()` would throw an exception too. [See documentation](https://docs.oracle.com/javase/7/docs/api/java/lang/reflect/AccessibleObject.html#setAccessible(boolean))

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,89 @@
...
+                            + "Class: " + implementation.getName()
+                            + ". Field: " + field.getName());
+                }
+            }
+        }
+        return clazzImplementationsInstance;
+    }
+
+    private Object instanceOf(Class<?> implementation) {
+        boolean present = implementation.isAnnotationPresent(Component.class);
```
nacenik, line 58: Do you really need this variable and this check? It will be always true, cause you have this check in the method "getImpl"

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,89 @@
...
+                     | InstantiationException | IllegalAccessException e) {
+                throw new RuntimeException("Can't create instance of " + implementation.getName());
+            }
+        }
+    }
+
+    private Class<?> getImpl(Class<?> clazz) {
+        if (!clazz.isAnnotationPresent(Component.class)
+                && !implementations.containsKey(clazz)) {
+            throw new RuntimeException("Unsupported class " + clazz.getName() + "!");
```
nacenik, line 80: Why this class is unsupported? Add more info to the exception method.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,89 @@
...
+        } else {
+            try {
+                Constructor<?> constructor = implementation.getConstructor();
+                Object instance = constructor.newInstance();
+                if (present) {
+                    components.put(implementation, instance);
+                }
+                return instance;
+            } catch (NoSuchMethodException | InvocationTargetException
+                     | InstantiationException | IllegalAccessException e) {
```
nacenik, line 71: You can put just `ReflectiveOperationException` instead of all exceptions that you wrote down. [See here](https://docs.oracle.com/javase/7/docs/api/java/lang/ReflectiveOperationException.html)

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,89 @@
...
+                throw new RuntimeException("Can't create instance of " + implementation.getName());
+            }
+        }
+    }
+
+    private Class<?> getImpl(Class<?> clazz) {
```
nacenik, line 77: Don't use the short form in method names. 
better to call this method `findImplementation`

nacenik (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+    }
+
+    private Class<?> findImplementation(Class<?> clazz) {
+        if (!clazz.isAnnotationPresent(Component.class)
+                && !implementations.containsKey(clazz)) {
+            throw new RuntimeException("Unsupported class " + clazz.getName() + "!"
+                    + "Didn't have supported implementation or @Component annotation!");
+        }
+        Class<?> impl = implementations.get(clazz);
+        if (impl != null) {
```
nacenik, line 68: For the future: don't forget, that you can check key using `map.containsKey()`

nacenik (approved):
LGTM! 🚀

PR #557, opened 2023-03-19
create Injector with @Component and @Inject

MaksymHorielikov:


Nikitos787 (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
+    private static final Class<Component> COMPONENT = Component.class;
+    private static final Class<Inject> INJECT = Inject.class;
```
nacenik, line 16: Why do you need these constants? Are they "magic"?

nacenik (changes requested):

nacenik (approved):
LGTM! 🚀

PR #559, opened 2023-03-19
Hw

PlagueEbola:


vika-trach (approved):
Very good job!

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        if (!(clazz.isAnnotationPresent(Component.class))) {
+            throw new RuntimeException("Can`t created instance of class."
+                    + " `@Component` annotation is missing");
```
KseniiaMakarova, line 32: which class?

KseniiaMakarova (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -28,8 +28,8 @@ public Object getInstance(Class<?> interfaceClazz) {
         Object clazzImplementationInstance = null;
         Class<?> clazz = findImplementation(interfaceClazz);
         if (!(clazz.isAnnotationPresent(Component.class))) {
-            throw new RuntimeException("Can`t created instance of class."
-                    + " `@Component` annotation is missing");
+            throw new RuntimeException("Can`t created instance of class " + clazz.getName()
+                    + ", `@Component` annotation is missing");
```
karma-o, line 32: A string could also be replaced by the `Component.class.getName()`

PR #566, opened 2023-03-19
[HW-DEPENDENCY-INJECTION] implement an injector that uses @component …

marianapolishchuk2169:
…and @inject annotations

### DIFF
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,13 +1,73 @@
 package mate.academy.lib;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.util.HashMap;
+import java.util.Map;
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private final Map<Class<?>, Object> instances = new HashMap<>();
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplamentationsIntance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        Field[] declaredFields = interfaceClazz.getDeclaredFields();
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Injection failed, "
+                    + "missing @Component annotaion on the class" + Component.class.getName());
+        }
+        for (Field field : declaredFields) {
+            if (field.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = getInstance(field.getType());
+                clazzImplamentationsIntance = createNewInstance(clazz);
+                try {
+                    field.setAccessible(true);
+                    field.set(clazzImplamentationsIntance, fieldInstance);
+                } catch (IllegalAccessException e) {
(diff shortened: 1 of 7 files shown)


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplamentationsIntance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        Field[] declaredFields = interfaceClazz.getDeclaredFields();
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Injection failed, "
+                    + "missing @Component annotaion on the class" + Component.class.getName());
+        }
```
potrap, line 29: you should write this 'if' on top of this method, because:
1. if class/interface hasn't `@Component` you will do redundant actions
2. imagine that interfaceClazz will be some interface which doesn't have Component annotation, in this case you put this in findImplementation method and there we'll get into this 'if':
`  if (interfaceClazz.isInterface()) {
        }`
which will return to us null (because we don't have such key), so
Class<?> clazz will be null and if we call some method to null we will have `NullPointerException();`

potrap (changes requested):

potrap (approved):
gj)

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+    private final Map<Class<?>, Object> instances = new HashMap<>();
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        if (!Objects.requireNonNull(clazz).isAnnotationPresent(Component.class)) {
```
sarakhmen, line 25: could throw non-informative NPE.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+            throw new RuntimeException("Injection failed, "
+                    + "missing @Component annotation on the class" + Component.class.getName());
+        }
+        Object clazzImplementationsInstance = null;
```
sarakhmen, line 29: I'd recommend checking if class implementation is present in instances map first. And return already present instance immediately, if present

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+        for (Field field : declaredFields) {
+            if (field.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = getInstance(field.getType());
+                clazzImplementationsInstance = createNewInstance(clazz);
+                try {
+                    field.setAccessible(true);
+                    field.set(clazzImplementationsInstance, fieldInstance);
+                } catch (IllegalAccessException e) {
+                    throw new RuntimeException("Can't initialize field. Class: "
+                            + clazz.getName() + ". Field: " + field.getName());
```
sarakhmen, line 40: let's also add an exception you’re catching in catch block to the RuntimeException object you throw, as stated in common mistakes

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
```
sarakhmen, line 60: let's also add an exception you’re catching in catch block to the RuntimeException object you throw, as stated in common mistakes

sarakhmen (changes requested):

PR #567, opened 2023-03-19
created Injector and added the annotations to the classes and all needed fields

ArtemGrunin:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of "
+                    + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = Map.of(
```
pechivo, line 67: You can make "interfaceImplementations" a class field and then in method you can only initialize it by "interfaceImplementations = Map.of(....)"

pechivo (approved):

src/main/java/mate.academy/lib/Injector.java
nacenik: Good point. Do you really need to create this map every time or can you reuse it?

nacenik (changes requested):

src/main/java/mate.academy/lib/Injector.java
ArtemGrunin: ok) thank you) You are right!)

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,78 @@
...
+                    + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        interfaceImplementations = Map.of(
+                ProductService.class, ProductServiceImpl.class,
+                ProductParser.class, ProductParserImpl.class,
+                FileReaderService.class, FileReaderServiceImpl.class
+        );
```
nacenik, line 72: Just create this map once. Do not create it every time

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,78 @@
...
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private final Map<Class<?>, Object> instances = new HashMap<>();
+    private Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
```
nacenik, line 17: make this field `final` too

nacenik (changes requested):
Fix some mistakes

src/main/java/mate.academy/lib/Injector.java
ArtemGrunin: finally refactored)

nacenik (approved):
LGTM! 🚀

PR #568, opened 2023-03-19
made an implementation of Injector

pechivo:


NadiaHuryk (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        interfaceImplementations = Map.of(FileReaderService.class, FileReaderServiceImpl.class,
+                ProductParser.class, ProductParserImpl.class, ProductService.class,
+                ProductServiceImpl.class);
```
nacenik, line 67: do you really need to create this map every time? 
You can initialize and fill it once using `Map.of()`

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Implementation of such class can't be created "
+                    + clazz.getName() + " . Check the presence of annotation");
```
nacenik, line 27: Which annotation? Do I need to add @SuppressWarnings or another?

nacenik (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private Map<Class<?>, Object> instances = new HashMap<>();
+    private Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
```
nacenik, line 17: it is a goo idea to make these field `final` too

nacenik (approved):
Good job) 🔥

PR #570, opened 2023-03-19
Implemented reflection API

AndrewkeY-hub:


KyslyiViacheslav (approved):
nice

src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Component {
```
nacenik, line 7: Do not forget to add @Target to custom annotations. See common mistakes

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+                 | InstantiationException | IllegalAccessException e) {
+            throw new RuntimeException("Can`t create new instance " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
```
nacenik, line 68: Make Interface Implementations map a class field. You can fill it in using Map.of().

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private Map<Class<?>, Object> instances = new HashMap<>();
```
nacenik, line 17: it is a good idea to make this field `final` too

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+
+    private Object createNewInstance(Class<?> clazz) {
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object o = constructor.newInstance();
+            if (!o.getClass().isAnnotationPresent(Component.class)) {
+                throw new RuntimeException("Class without annotation @Component");
```
nacenik, line 54: Which class? Add more info to the exception message

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        Field[] declaredFields = clazz.getDeclaredFields();
+        for (Field field : declaredFields) {
+            if (field.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = getInstance(field.getType());
+                clazzImplInstance = createNewInstance(clazz);
+                field.setAccessible(true);
```
nacenik, line 31: put this line inside the try block, cause `setAccessible()` would throw an exception too. [See documentation](https://docs.oracle.com/javase/7/docs/api/java/lang/reflect/AccessibleObject.html#setAccessible(boolean))

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+                try {
+                    field.set(clazzImplInstance, fieldInstance);
+                } catch (IllegalAccessException e) {
+                    throw new RuntimeException("Can`t initialize field value "
+                            + field.getName());
```
nacenik, line 36: In which class? The same field could be in different classes

nacenik (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,80 @@
...
+                 | InstantiationException | IllegalAccessException e) {
+            throw new RuntimeException("Can`t create new instance " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = Map.of(
+                FileReaderService.class, FileReaderServiceImpl.class,
+                ProductParser.class, ProductParserImpl.class,
+                ProductService.class, ProductServiceImpl.class);
```
nacenik, line 74: Make Interface Implementations map a class field.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,80 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object o = constructor.newInstance();
+            if (!o.getClass().isAnnotationPresent(Component.class)) {
+                throw new RuntimeException(o.getClass().getName()
+                        + " class without @Component annotation");
+            }
```
nacenik, line 61: I'd recommend you check the annotation before you get and call a class constructor

nacenik (changes requested):

src/main/java/mate.academy/lib/Injector.java
AndrewkeY-hub: Could you explain to me where is a mistake?

nacenik: initialize it inline. Instead of `new HashMap()` make `Map.of()`

maxline (approved):
Good job! Approved

PR #575, opened 2023-03-20
Implemented solution for hw.

RandomEastEcho:


KyrylH (approved):

nacenik (changes requested):

PR #576, opened 2023-03-20
Maksym Yashyn Homework

Nimakel:


Deschna (changes requested):
Overall good, but two small fixes

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -36,7 +36,7 @@ public Object getInstance(Class<?> interfaceClazz) {
                     field.setAccessible(true);
                     field.set(clazzImplementation, fieldInstance);
                 } catch (IllegalAccessException e) {
-                    throw new RuntimeException("Can`t initialize filed value. Class:"
+                    throw new RuntimeException("Can`t initialize field value. Class:"
                             + clazz.getName() + ", Filed: " + field.getName());
```
Deschna, line 40: One more typo here, but im think u can fix this after approve)

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+    );
+    private Map<Class<?>, Object> instances = new HashMap<>();
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementation = null;
```
Andriy-dz, line 28: I'd recommend checking if class implementation is present in instances map first. And return the already present instance immediately, if present

Andriy-dz (changes requested):
Very goog solution! Let`s improve it)

PR #578, opened 2023-03-20
Dependency injection practice. Implemented Injector class

Deschna:


Nimakel (changes requested):
Good job. Check my comments and fix this problems

src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,12 @@
 package mate.academy.lib;
 
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(value = {ElementType.TYPE})
```
maxline, line 9: can be simplified as 
```
@Target(ElementType.TYPE)
```

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,12 @@
 package mate.academy.lib;
 
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(value = {ElementType.FIELD})
```
maxline, line 9: `@Target(ElementType.TYPE)`

maxline (approved):
Good job! Approved with couple small remarks, please take a look for future improvements.

PR #581, opened 2023-03-20
added implementation

IvanZuravlev228:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+                 | InvocationTargetException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementation = new HashMap<>();
+        interfaceImplementation.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementation.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementation.put(ProductService.class, ProductServiceImpl.class);
```
Andriy-dz, line 69: you create a map each time. Perhaps it is better to make it a class field and initialize it in a static block?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InstantiationException | IllegalAccessException
+                 | InvocationTargetException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
```
Andriy-dz, line 61: add another ", e" at the end. Likewise in similar cases

Andriy-dz (changes requested):

karma-o (approved):

PR #583, opened 2023-03-20
Implemented custom Injector

ArtemPonomar:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,88 @@
...
+    private static final String NO_ANNOTATION
+            = "Could not find class of %s annotated with @Component";
+    private static final String CREATION_EXCEPTION = "Could not create instance of class ";
+    private static final String FIELD_INIT_EXCEPTION = "Can`t initialize field ";
+
+    private Injector() {
+        implementations = new HashMap<>();
+        implementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        implementations.put(ProductParser.class, ProductParserImpl.class);
+        implementations.put(ProductService.class, ProductServiceImpl.class);
```
redmanO-o, line 30: recomendation from CM:
"Make Interface Implementations map a class field. You can fill it in using Map.of()."

redmanO-o (approved):
GREAT!!!

maxline (changes requested):
Good job! let's improve it

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,80 @@
...
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Class<?> impl = interfaceClazz;
+        if (implementations.containsKey(interfaceClazz)) {
+            impl = implementations.get(interfaceClazz);
+        }
+        if (!impl.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException(
+                    String.format(NO_ANNOTATION, interfaceClazz.getName()));
+        }
```
nacenik, line 77: To avoid using one more variable, you can put this check first, and use just `interfaceClazz` variable

nacenik (approved):
Great job! You can improve your solution if you want 😉

PR #585, opened 2023-03-21
implemented injector class, created Injector and Component annotation…

Skrudj:
…s, set annotations above the fields and classes

### DIFF
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,13 +1,77 @@
 package mate.academy.lib;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.util.HashMap;
+import java.util.Map;
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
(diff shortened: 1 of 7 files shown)


PlagueEbola (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+            return instances.get(clazz);
+        }
+        Constructor<?> constructor = null;
+        try {
+            constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                 | InstantiationException | IllegalAccessException e) {
```
sarakhmen, line 73: as stated in common mistakes, it is better to replace many exceptions that have a common parent with a general parental exception

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+        return interfaceClazz;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        Constructor<?> constructor = null;
+        try {
+            constructor = clazz.getConstructor();
```
sarakhmen, line 68:
```suggestion
        try {
            Constructor<?> constructor = clazz.getConstructor();
```

sarakhmen (changes requested):

Andriy-dz (approved):
Good Job!

PR #586, opened 2023-03-24
 implement an injector that uses @Component and @Inject annotations t…

Yulia-Kas:
…o initialize and return Object instances

### DIFF
--- a/src/main/java/mate.academy/lib/Component.java
+++ b/src/main/java/mate.academy/lib/Component.java
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Component {
 
 }
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,13 +1,75 @@
 package mate.academy.lib;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.util.HashMap;
+import java.util.Map;
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
-    private static final Injector injector = new Injector();
+    private static final Injector INJECTOR = new Injector();
+    private static final Map<Class<?>, Object> INSTANCES = new HashMap<>();
 
     public static Injector getInjector() {
-        return injector;
+        return INJECTOR;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        if (clazz.isAnnotationPresent(Component.class)) {
+            Field[] declaredFields = clazz.getDeclaredFields();
+            for (Field field : declaredFields) {
+                if (field.isAnnotationPresent(Inject.class)) {
+                    Object fieldInstance = getInstance(field.getType());
+                    clazzImplementationInstance = createNewInstance(clazz);
+                    try {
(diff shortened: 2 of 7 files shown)


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Component {
```

Rommelua (changes requested):

Before you create PR with your homework, you need to go through the checklist under the task and correct **all the points** described there. The mentor will not check the work until the checklist points are corrected

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
-    private static final Injector injector = new Injector();
+    private static final Injector INJECTOR = new Injector();
```
andriy-shulepa, line 16: it is not constant because it's created object and after each start of application there will be different instance

so, don't use constant naming convention here

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+        if (INSTANCES.containsKey(clazz)) {
+            return INSTANCES.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            INSTANCES.put(clazz, instance);
+            return instance;
+        } catch (InstantiationException | IllegalAccessException
+                | InvocationTargetException | NoSuchMethodException e) {
```
andriy-shulepa, line 60: you can catch just ReflectiveOperationException

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+                        field.setAccessible(true);
+                        field.set(clazzImplementationInstance, fieldInstance);
+                    } catch (IllegalAccessException e) {
+                        throw new RuntimeException("Can't initialize field value. "
+                                + "Class: " + clazz.getName() + ". Field: " + field.getName());
+                    }
+                }
+            }
+        } else {
+            throw new RuntimeException("There is no annotation Component in class " + clazz);
```
andriy-shulepa, line 42:
```suggestion
            throw new RuntimeException("There is no annotation Component in class " + clazz.getName());
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+                | InvocationTargetException | NoSuchMethodException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementation = new HashMap<>();
+        interfaceImplementation.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementation.put(ProductService.class, ProductServiceImpl.class);
+        interfaceImplementation.put(ProductParser.class, ProductParserImpl.class);
```
andriy-shulepa, line 69: move this map to class level to avoid creating each time findImplementation is called

andriy-shulepa (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        interfaceImplementation.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementation.put(ProductService.class, ProductServiceImpl.class);
+        interfaceImplementation.put(ProductParser.class, ProductParserImpl.class);
```
aaananas, line 68: put these values when initializing the map field - use Map.of method

aaananas (approved):

PR #588, opened 2023-03-28
solved task - implemented Injector

yaaariiik:


olekskov (changes requested):
Please go to Main class and perform instruction in comments for testing.
I have doubt that this solution is working as is.
Please laso include Main class to next commit

Ivan95kos (approved):
Great job

PR #591, opened 2023-03-30
Hw dependency injection

vromanyuk:


src/main/java/mate.academy/service/impl/ProductParserImpl.java
```diff
@@ -1,20 +1,24 @@
...
 
+@Component
 public class ProductParserImpl implements ProductParser {
     public static final int ID_POSITION = 0;
     public static final int NAME_POSITION = 1;
     public static final int CATEGORY_POSITION = 2;
     public static final int DESCRIPTION_POSITION = 3;
     public static final int PRICE_POSITION = 4;
+    @Inject
+    private Product product;
```
olekskov, line 17: Injector is aimed to substitute interface with it's implementation.
Product is a pure model which doesn't implement any interface so cannot and shouldn't be injected

src/main/java/mate.academy/model/Product.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.model;
 
 import java.math.BigDecimal;
+import mate.academy.lib.Component;
 
+@Component
```
olekskov, line 6: It's not a component

olekskov (changes requested):

src/main/java/mate.academy/model/Product.java
vromanyuk: Thank you a lot, I understand

PR #593, opened 2023-04-01
Solution was added

Natalia-Shcheglova:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,78 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                | InstantiationException | IllegalAccessException e) {
```
aaananas, line 67: those exceptions extend the same class - use it

aaananas (changes requested):

Rommelua (approved):

PR #594, opened 2023-04-02
initial commit

Rolik-Oleksandr:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Component {
```
olekskov, line 7: Do not forget to add @Target to custom annotations.

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```

olekskov (changes requested):
In addition to comments below please also check main class and perform actions advised there so we can see you know how to use injector

src/main/java/mate.academy/lib/Component.java
okuzan: target: only TYPE. Should it be possible to annotate local variables as a component?)

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,12 @@
 package mate.academy.lib;
 
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(value = {ElementType.FIELD, ElementType.LOCAL_VARIABLE})
```
okuzan, line 8: only field

okuzan (changes requested):
Add main class

kozhukhovsky (changes requested):
Add main class

okuzan (approved):
Good job

PR #597, opened 2023-04-06
added proper solution for this task

polikarpov13:


src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Inject {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Inject {
```
kozhukhovsky, line 7: It would be great to add a Target annotation here to indicate that we will put our own annotation only on fields

kozhukhovsky (approved):

PR #601, opened 2023-04-19
hw-17-solution

IhorPetrytskyi333:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
-public @interface Component {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Component {
 }
+
```
kozhukhovsky, line 9: remove empty line

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
-public @interface Inject {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
 
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Inject {
```
kozhukhovsky, line 8: It would be great to add a Target annotation here to indicate that we will put our own annotation only on fields

src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
-public @interface Component {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Component {
```
kozhukhovsky, line 7: It would be great to add a Target annotation here to indicate that we will put our own annotation only on class

kozhukhovsky (approved):

PR #606, opened 2023-04-25
Implemented an injector that uses @Component and @Inject annotation;

SerhiUA:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        interfaceImplementation.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementation.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementation.put(ProductService.class, ProductServiceImpl.class);
```
okuzan, line 26: use Map.of() on line 16 instead of setting values here

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+
+        if (!findImplementation(interfaceClazz).isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Injection failed, missing @Component "
+                    + "annotation on the class " + interfaceClazz);
+        }
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        Field[] declaredField = clazz.getDeclaredFields();
+        for (Field field: declaredField) {
```
okuzan, line 35:
```suggestion
        for (Field field : declaredField) {
```

okuzan (changes requested):

PR #610, opened 2023-05-10
the Injector class is implemented and annotations are added to the im…

0-zagar-0:
…plementation classes

### DIFF
--- a/src/main/java/mate.academy/lib/Component.java
+++ b/src/main/java/mate.academy/lib/Component.java
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Component {
 
 }
--- a/src/main/java/mate.academy/lib/Inject.java
+++ b/src/main/java/mate.academy/lib/Inject.java
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Inject {
 
 }
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,13 +1,79 @@
 package mate.academy.lib;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.util.HashMap;
+import java.util.Map;
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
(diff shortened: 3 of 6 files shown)


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
okuzan, line 6: add Target

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
okuzan, line 6: add Target

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,78 @@
...
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance " + clazz.getName(), e);
+        }
+    }
+    
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementation = new HashMap<>();
+        interfaceImplementation.put(ProductService.class, ProductServiceImpl.class);
+        interfaceImplementation.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementation.put(FileReaderService.class, FileReaderServiceImpl.class);
```
okuzan, line 67: let;s make it a class field
    private static final Map<Class<?>, Class<?>> interfaceImplementationMap = Map.of(
    );

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,78 @@
...
+        
+        if (interfaceClazz.isInterface()
+                && interfaceImplementation.get(interfaceClazz)
+                .isAnnotationPresent(Component.class)) {
```
okuzan, line 71: you don't need to check it here
here if interfaceClazz is an interface, find an implementation, if not - return what you get from input

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,78 @@
...
     private static final Injector injector = new Injector();
+    private final Map<Class<?>, Object> instances = new HashMap<>();
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Class<?> clazz = findImplementation(interfaceClazz);
```
okuzan, line 23: here check if your clazz has @Component annotation

okuzan (changes requested):

olekskov (approved):
Good job

PR #612, opened 2023-05-11
solved task

AVykhovanok:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,78 @@
...
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                        | InstantiationException | IllegalAccessException e) {
+            throw new RuntimeException(String.format(
+                    "Can`t create instance of class - %s", clazz.getName()));
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
```
olekskov, line 68: Consider moving the 'interfaceImplementaionMap' to a class-level field instead of recreating it on every call to 'findImplementation'. This will improve performance.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,78 @@
...
+        Map<Class<?>, Class<?>> interfaceImplementaionMap = new HashMap<>();
+        interfaceImplementaionMap.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementaionMap.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementaionMap.put(ProductService.class, ProductServiceImpl.class);
+        if (interfaceClazz.isInterface()) {
+            return interfaceImplementaionMap.get(interfaceClazz);
```
olekskov, line 74: You should check if the implementation class is present in the 'interfaceImplementaionMap' for the given interfaceClazz. If not, throw a suitable exception.

Ivan95kos (approved):
Great job

PR #613, opened 2023-05-12
completed task

sZlishchev:


OlehPryshchepa (approved):
great job:)

src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
maxline, line 6: please also add `Target` with proper type

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,84 @@
...
+                }
+            }
+        }
+        if (clazzImplementationInstance == null) {
+            clazzImplementationInstance = createNewInstance(clazz);
+        }
+        return clazzImplementationInstance;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
```
maxline, line 45:
```suggestion
    private Object getOrCreateNewInstance(Class<?> clazz) {
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,84 @@
...
+
+    private void checkComponentAnnotation(Class<?> clazz) {
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Class " + clazz.getName()
+                    + " doesn't have " + Component.class + " annotation");
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = Map.of(
```
maxline, line 68: should we create the map every time when findImplementation() method called?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,84 @@
...
+                ProductService.class, ProductServiceImpl.class,
+                ProductParser.class, ProductParserImpl.class,
+                FileReaderService.class, FileReaderServiceImpl.class
+        );
+        if (interfaceClazz.isInterface()) {
+            Class<?> implementation = interfaceImplementations.get(interfaceClazz);
```
maxline, line 74: what will happen if the implementation is not found?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,84 @@
...
+            if (implementation.isAnnotationPresent(Component.class)) {
```
maxline, line 75: could we reuse `checkComponentAnnotation()` to avoid the code duplication?

maxline (changes requested):
Good job! let's improve the solution

src/main/java/mate.academy/lib/Injector.java
maxline: chek if `implementation == null` and throw an exception

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,80 @@
...
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Class " + clazz.getName()
+                    + " doesn't have " + Component.class + " annotation");
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (interfaceClazz.isInterface()) {
+            Class<?> implementation = interfaceImplementations.get(interfaceClazz);
+            checkComponentAnnotation(interfaceClazz);
```
maxline, line 74: move out this  `checkComponentAnnotation(interfaceClazz)` from `findImplementation()` and put it after this method, and check class that is  result of `findImplementation()` method.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,80 @@
...
+            return implementation;
+        } else {
+            throw new RuntimeException("Can't find implementation for:" + interfaceClazz.getName());
```
maxline, line 77:
```suggestion
            return interfaceClazz;
```

maxline (changes requested):
Good job! Left just a couple more comments

maxline (approved):
Well done! Approved

PR #614, opened 2023-05-12
HW-jv-dependency-injection-12-Of-May

ViktoriiaSvidovska:


Ivan95kos (approved):
For the future, do not write changes in the master branch, create a separate branch for this

PR #621, opened 2023-05-20
Jv dependency injection hw

TymurProkhorov:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Component {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Component {
```
olekskov, line 7: Add the @Target(ElementType.TYPE) annotation to specify that the @Component annotation can only be applied to classes.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,78 @@
...
+            return instance;
+        } catch (NoSuchMethodException
+                 | InvocationTargetException
+                 | InstantiationException
+                 | IllegalAccessException e) {
+            throw new RuntimeException("Can`t create new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
```
olekskov, line 71: With such approach you will recreate the map for every call of getInstance method.
Please move map initialization to class level to be performed only once

aaananas (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can`t create new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
```
olekskov, line 71: This will be repeated each time you call findImplementation method. It would be better to performed only once

olekskov (approved):
Good job overall, approved but for further - filling map with values that are not going to change is better to do once

PR #622, opened 2023-05-22
task solved

MykytaKat:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        if (interfaceClazz.isInterface()) {
+            return interfaceImplementations.get(interfaceClazz);
```
maxline, line 71: let's also throw an runtime exception if the implementation for the interface was not found

maxline (changes requested):
Good job! let's improve the solution

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+                }
+            }
+        }
+        if (clazzImplementationInstance == null) {
+            clazzImplementationInstance = createNewInstance(clazz);
+        }
+        return clazzImplementationInstance;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
```
maxline, line 50:
```suggestion
    private Object getOrCreateNewInstance(Class<?> clazz) {
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                 | InstantiationException | IllegalAccessException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
```
maxline, line 66: map is created every time when method `findImlementation()` called

maxline (changes requested):
A couple additional remarks added.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
```
maxline, line 67: please also put map these initialisation lines to the class level, where map variable is introduced. we don't need fill map up with values every time when the method findImplementation() called.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+        if (interfaceClazz.isInterface()) {
+            return interfaceImplementations.get(interfaceClazz);
+        }
+        throw new RuntimeException("The implementation for the interface"
```
maxline, line 71: this exception should be thrown only if 
`interfaceImplementations.get(interfaceClazz) == null` inside of the block `if (interfaceClazz.isInterface()) {`

maxline (changes requested):
Well done! Left just a couple more remarks.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,83 @@
...
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (interfaceClazz.isInterface()) {
+            return interfaceImplementations.get(interfaceClazz);
+        }
+        if (interfaceImplementations.get(interfaceClazz) == null) {
+            throw new RuntimeException("The implementation for the interface"
+                    + interfaceClazz.getName() + " was not found.");
+        }
```
maxline, line 80:
```suggestion
          if (interfaceImplementations.get(interfaceClazz) == null) {
               throw new RuntimeException("The implementation for the interface"
                    + interfaceClazz.getName() + " was not found.");
           }
            return interfaceImplementations.get(interfaceClazz);
        }
```

maxline (approved):
Well done! Approved with one remark. You need 1) validate if value ==null 2) throw exception and after that 3) return value. In your implementation step 2 and 3 was swapped.

PR #623, opened 2023-05-25
Implement injector used annotation - component and inject

Gamemod13:


src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,11 @@
 package mate.academy.lib;
 
-public @interface Inject {
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
 
+@Target(ElementType.FIELD)
+@Retention(RetentionPolicy.RUNTIME)
```
MykytaKat, line 9: Retention should be the first for better readability

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,78 @@
...
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> implementations = Map.of(
+                ProductParser.class, ProductParserImpl.class,
+                FileReaderService.class, FileReaderServiceImpl.class,
+                ProductService.class, ProductServiceImpl.class
+        );
+        if (interfaceClazz.isInterface()) {
+            return implementations.get(interfaceClazz);
```
MykytaKat, line 74: let's throw an runtime exception if the implementation for the interface was not found

MykytaKat (changes requested):
good job! let's improve your solution

MykytaKat (approved):
good job

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,84 @@
...
+    private Object createNewInstance(Class<?> clazz) {
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | IllegalAccessException
```
maxline, line 61: from common mistakes: "It is better to replace many exceptions that have a common parent with a general parental exception."

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,84 @@
...
+            throw new RuntimeException("Can't create instance of " + clazz.getSimpleName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> implementations = Map.of(
+                ProductParser.class, ProductParserImpl.class,
+                FileReaderService.class, FileReaderServiceImpl.class,
+                ProductService.class, ProductServiceImpl.class
+        );
```
maxline, line 72: should we create and fill the map every time when method `findImplemention()` are called?

maxline (changes requested):
Good job! Let's improve the solution

sarakhmen (approved):
Good job :)

PR #630, opened 2023-05-31
Created the injector and implemented the necessary methods for the co…

V-Solianyk:
…rrect operation of the injector, added the necessary annotations over the fields and classes.

### DIFF
--- a/src/main/java/mate.academy/lib/Inject.java
+++ b/src/main/java/mate.academy/lib/Inject.java
@@ -1,5 +1,11 @@
 package mate.academy.lib;
 
-public @interface Inject {
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
 
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+public @interface Inject {
 }
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,13 +1,73 @@
 package mate.academy.lib;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.util.HashMap;
+import java.util.Map;
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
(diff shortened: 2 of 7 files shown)


src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,11 @@
...
-public @interface Inject {
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
 
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+public @interface Inject {
 }
```
Tupadzaky, line 11: Same annotation issue here

Tupadzaky (changes requested):
The only problem is that interfacesImplementations Map in your injector should be declared on class level.
I don't think that annotations order in your custom annotations are a big deal

Tupadzaky (approved):
Good job!

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+                }
+            }
+        }
+        if (clazzImplementationInstance == null) {
+            clazzImplementationInstance = createNewInstance(clazz);
+        }
+        return clazzImplementationInstance;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
```
maxline, line 54:
```suggestion
    private Object gerOrreateNewInstance(Class<?> clazz) {
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can not create a new instance of " + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (interfaceClazz.isInterface()) {
+            return interfacesImplementation.get(interfaceClazz);
```
maxline, line 70: let's also throw an runtime exception if the implementation for the interface was not found, e.g.:
```
           }
```

maxline (changes requested):
Good job! Let's improve the solution

src/main/java/mate.academy/lib/Injector.java
V-Solianyk: I just needed to change the name of the method here, but maybe it would be more correct to call it like this:
"getOrCreateNewInstance" ?

maxline (approved):
Well done! Approved

PR #631, opened 2023-06-02
Solution

MTM1963:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                 | InstantiationException | IllegalAccessException e) {
```
aaananas, line 65: those exceptions have the same parent class - use it instead

aaananas (approved):

src/main/java/mate.academy/lib/Injector.java
MTM1963: Okay

PR #632, opened 2023-06-02
Add all needed changes to Injector class and other classes

ykhmaruk:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can`t create a new instance of " + clazz.getName());
```
volodymyrruso, line 60: The code catches ReflectiveOperationException, but throws it as a RuntimeException

volodymyrruso (changes requested):
great job, let's improve you solution a bit

volodymyrruso (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+                }
+            }
+        }
+        if (clazzImplementationInstance == null) {
+            clazzImplementationInstance = createNewInstance(clazz);
+        }
+        return clazzImplementationInstance;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
```
maxline, line 50:
```suggestion
    private Object getOrCreateNewInstance(Class<?> clazz) {
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can`t create a new instance of " + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        interfaceImplementation = Map.of(
```
maxline, line 65: Let's move creation of map to the class level (at the begin of the class). It is not necessary to create it every time when the method is called with the same values.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can`t create a new instance of " + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (interfaceImplementations.get(interfaceClazz) == null) {
```
maxline, line 68: move this check `(interfaceImplementations.get(interfaceClazz) == null) `
 inside of  `if (interfaceClazz.isInterface()) {` block

maxline (changes requested):
One more minor comment

sarakhmen (approved):
Good job :)

PR #634, opened 2023-06-03
tasks done

marimarig:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,82 @@
...
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (interfaceClazz.isInterface()) {
+            Optional<Class<?>> implementation = Optional
+                    .ofNullable(INTERFACE_IMPLEMENTATIONS.get(interfaceClazz));
+            return implementation.orElseThrow(() -> new RuntimeException("Implementation for: "
+                    + interfaceClazz.getName() + " is null."));
+        }
+        return interfaceClazz;
```
chrisnyrka, line 80: I think, null check is unnecessary, because your map is a constant and will not change with the progress of the program

chrisnyrka (approved):
Good job!)

src/main/java/mate.academy/lib/Injector.java
maxline: The Maryna's implementation is correct. The possible case in future - the new class with annotation usage added, but map is be forgotten to update.

maxline (changes requested):
Well done! Left one minor remark and one comment for student review.

maxline (approved):
Well done! Approved

PR #635, opened 2023-06-04
implemented task

chrisnyrka:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,79 @@
...
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private Map<Class<?>, Object> instances = new HashMap<>();
+
+    private final Map<Class<?>, Class<?>> interfaceImplementations = Map.of(
```
marimarig, line 18: I think nedd to add "static"

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,79 @@
...
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (interfaceClazz.isInterface()) {
+            return interfaceImplementations.get(interfaceClazz);
+        }
+        return interfaceClazz;
```
marimarig, line 77: Maybe need to add check if interfaceImplementations == null

marimarig (approved):
Good job)

src/main/java/mate.academy/lib/Injector.java
chrisnyrka: I think it's unnecessary if the map will be a constant

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+                }
+            }
+        }
+        if (clazzImplementationInstance == null) {
+            clazzImplementationInstance = createNewInstance(clazz);
+        }
+        return clazzImplementationInstance;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
```
maxline, line 54:
```suggestion
    private Object getOrCreateNewInstance(Class<?> clazz) {
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (interfaceClazz.isInterface()) {
+            return INTERFACE_IMPLEMENTATIONS.get(interfaceClazz);
```
maxline, line 70: let's also throw an runtime exception if the implementation for the interface was not found, e.g.:
```
           }
```

maxline (changes requested):
Good job! Left a couple of minor comments

maxline (approved):
Well done! Approved

PR #643, opened 2023-06-06
add di solution

AnastasiiaVariuchuk:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,82 @@
...
+import java.util.Map;
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
+    private static final Map<Class<?>, Class<?>> interfaceImplementations = Map.of(
```
ArinaPolna, line 14: Your code is clean and informative, excellent job) just a small remark: you can move this interfaceImplementations Map to findImplementation method as was described in video (cause we gonna use this map only in this method)

ArinaPolna (changes requested):

ArinaPolna (approved):

src/main/java/mate.academy/lib/Injector.java
AnastasiiaVariuchuk: I think it's more convenient to leave it in this way)

maxline: this implementation is ok.

maxline (approved):
Perfect solution! Approved

PR #645, opened 2023-06-06
done task

ViktoriiaSemenchenko:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+            throw new RuntimeException("Can't create a new instance of "
+                    + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (interfaceClazz.isInterface()) {
+            return interfaceImplementations.get(interfaceClazz);
+        }
+        return interfaceClazz;
```
OlehPryshchepa, line 75: let's throw a runtime exception if the class was not found.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+                }
+            }
+        }
+        if (clazzImplementationInstance == null) {
+            clazzImplementationInstance = createNewInstance(clazz);
+        }
+        return clazzImplementationInstance;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
```
OlehPryshchepa, line 56: raname this method to "getOrCreateNewInstance"

OlehPryshchepa (changes requested):
good job, let`s improve your solution

OlehPryshchepa (approved):
Great Job! keep pushing!

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,78 @@
...
+                    + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (interfaceClazz.isInterface()) {
+            return interfaceImplementations.get(interfaceClazz);
+        }
+        throw new RuntimeException("Can`t find implementation class for interface: "
+                + interfaceClazz.getName());
```
sarakhmen, line 76: Here I'd recommend following the next logic: if clazz is an interface, than get from interfaceClazz map and check if received value is null (and if yes - throw an expection), if clazz isn't an interface, than simply return it

sarakhmen (approved):
Good job, you can improve your solution if you want ;)

PR #646, opened 2023-06-06
main() work, tests done, checkstyle done

MykNich:


ASnazar (approved):
Good job! I didn't find a mistake, well done

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,82 @@
...
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (interfaceImplementations.get(interfaceClazz) == null) {
+            throw new RuntimeException("Implementation for the interface"
+                    + interfaceClazz.getName() + " was not found");
+        }
```
sarakhmen, line 76: let's move this 'if' into the next one, as we have to get implementation from this map only in case clazz is an interface

PR #648, opened 2023-06-06
done

vladyslavkryvets:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,83 @@
...
+        if (interfaceClazz.isInterface()) {
+            return interfaceImplementations.get(interfaceClazz);
+        }
+        if (interfaceImplementations.get(interfaceClazz) == null) {
+            throw new RuntimeException("The implementation for the interface"
+                    + interfaceClazz.getName() + " was not found.");
+        }
+        return interfaceClazz;
     }
 }
```
Valdemar619, line 83: I`m not sure but I think that you need a empty line.

Valdemar619 (approved):
Good job just a little fix

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,83 @@
...
+    private static final Map<Class<?>, Class<?>> interfaceImplementations;
+
+    static {
+        injector = new Injector();
+        instances = new HashMap<>();
+        interfaceImplementations = new HashMap<>();
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+    }
```
sarakhmen, line 26: there is no need to make your instances and this 'if' should be moved into the previous one a static field. Also consider initializing this 'if' should be moved into the previous one using Map.of().

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,83 @@
...
+                }
+            }
+        }
+        if (clazzImplementationInstance == null) {
+            clazzImplementationInstance = createNewInstance(clazz);
+        }
+        return clazzImplementationInstance;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
```
sarakhmen, line 59:
```suggestion
    private Object getOrCreateInstance(Class<?> clazz) {
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,83 @@
...
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (interfaceClazz.isInterface()) {
+            return interfaceImplementations.get(interfaceClazz);
+        }
+        if (interfaceImplementations.get(interfaceClazz) == null) {
+            throw new RuntimeException("The implementation for the interface"
+                    + interfaceClazz.getName() + " was not found.");
+        }
```
sarakhmen, line 80: this 'if' should be moved into the previous one

sarakhmen (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,82 @@
...
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Class<?> implementationClazz = interfaceImplementations.get(interfaceClazz);
+        if (implementationClazz == null) {
+            throw new RuntimeException("The implementation for the interface "
+                    + interfaceClazz.getName() + " wasn`t found.");
+        }
+        if (implementationClazz.isInterface()) {
+            return findImplementation(implementationClazz);
+        }
+        return implementationClazz;
```
sarakhmen, line 80:
```suggestion
        if (interfaceClazz.isInterface()) {
            Class<?> implementationClazz = interfaceImplementations.get(interfaceClazz);
            if (implementationClazz == null) {
                throw new RuntimeException("The implementation for the interface"
                        + interfaceClazz.getName() + " was not found.");
            }
            return implementationClazz;
        }
        return interfaceClazz
```

sarakhmen (changes requested):
Let's apply some changes one more time. Please review my suggestion

PR #650, opened 2023-06-06
Implement injector pass all tests

Fafakepa:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        if (!clazz.isAnnotationPresent(Component.class)) {
```
mykyta-udav, line 28: Maybe create method for this if

mykyta-udav (changes requested):
Let's improve your solution

rovein (approved):
Great job! Check one minor comment.

PR #651, opened 2023-06-06
Added task implementation

pkobevko:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,71 @@
...
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = Map.of(
+                FileReaderService.class, FileReaderServiceImpl.class,
+                ProductParser.class, ProductParserImpl.class,
+                ProductService.class, ProductServiceImpl.class
+        );
```
sarakhmen, line 68: Let's move creation of map to the class level (at the begin of the class). It is not necessary to create it every time when the method is called with the same values. Also I'd recommend checking whether clazz is interface (and then throwing an exception if you get null from map) or not (and than return interfaceClazz directly)

sarakhmen (changes requested):

rovein (approved):
Great job, well done)

PR #654, opened 2023-06-06
implemented an injector that uses @Component and @Inject annotations

m-pnsk:


DV7Art:
I think you have an interesting implementation.

DV7Art (approved):

rovein (approved):
Great job!

PR #657, opened 2023-06-08
added solution for Dependency injection task

Fiasko3304:


src/main/java/mate.academy/Main.java
maksigors, line 1: good job!

src/main/java/mate.academy/Main.java
```diff
@@ -10,7 +10,7 @@ public class Main {
     public static void main(String[] args) {
         // Please test your Injector here. Feel free to push this class as a part of your solution
```
maksigors, line 11: you need to remove this line description

sarakhmen: agree with @maksigors , let's remove redundant comments

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+            if (clazzImpl.isAnnotationPresent(Component.class)) {
+                return clazzImpl;
+            }
+        }
+        if (interfaceClazz.isAnnotationPresent(Component.class)) {
+            return interfaceClazz;
+        }
+        throw new RuntimeException("Can't create instance of Class: " + clazzImpl.getName()
+                + " - is not annotated with @Component");
     }
```
sarakhmen, line 76: here you have a duplication of 'if' block, consider optimizing your code. For example, you can move .isAnnotationPresent() check outside of this method

sarakhmen (changes requested):
Let's slightly improve your solution ;)

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+    }
+
+    private Class<?> findImplementationClass(Class<?> interfaceClazz) {
+        Class<?> clazzImpl = interfaceClazz;
+        if (interfaceClazz.isInterface() && interfaceImplementations.containsKey(interfaceClazz)) {
+            clazzImpl = interfaceImplementations.get(interfaceClazz);
+        }
+        if (clazzImpl.isAnnotationPresent(Component.class)) {
+            return clazzImpl;
+        }
```
sarakhmen, line 70: what is interfaceClazz is interface, annotated with Component, but isn't present in your implementations map? In this case interface will be returned from the method, which is incorrect behaviour

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+        throw new RuntimeException("Can't create instance of Class: " + clazzImpl.getName()
+                + " - is not annotated with @Component");
```
sarakhmen, line 72: let's check on isAnnotationPresent in your getInstance method right after findImplementationClass() method is called

sarakhmen (changes requested):
Let's improve your solution one more time :)

sarakhmen (approved):
Good job :)

PR #663, opened 2023-06-16
implemented Injector class with Component and Inject annotations

vaniii1:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Component {
```

Rommelua (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,71 @@
...
+    private Map<Class<?>, Object> instances = new HashMap<>();
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = interfaceImplementations.get(interfaceClazz);
```
okuzan, line 28: create method `private Class<?> findImplementation(Class<?> interfaceClazz) {` 
if interface class is interface, get from map, if not, return itself

okuzan (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,78 @@
...
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create an instance of the class " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = Map.of(
+                ProductService.class, ProductServiceImpl.class,
+                ProductParser.class, ProductParserImpl.class,
+                FileReaderService.class, FileReaderServiceImpl.class);
```
okuzan, line 64: this should be a class field. Now you create a map each time a method is called

okuzan (changes requested):

okuzan (approved):
Well done ❤️

PR #665, opened 2023-06-18
added an injector that uses @Component and @Inject annotations and @T…

TemkaHillfinger:
…arget annotations

### DIFF
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,13 +1,74 @@
 package mate.academy.lib;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.util.HashMap;
+import java.util.Map;
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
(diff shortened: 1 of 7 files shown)


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementetion(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImpl = new HashMap<>();
+        interfaceImpl.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImpl.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImpl.put(ProductService.class, ProductServiceImpl.class);
```
okuzan, line 68: let's create this map only once instead of per each method call - make it a class field. Initialize at declaration using Map.of

okuzan (changes requested):

aaananas (approved):

aaananas (approved):

PR #669, opened 2023-06-23
jv-DI

s-hrynus:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Component {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Component {
```
Ivan95kos, line 7: Add https://github.com/target here

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.isInterface());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations
+                = Map.of(FileReaderService.class, FileReaderServiceImpl.class,
+                        ProductParser.class, ProductParserImpl.class,
+                        ProductService.class, ProductServiceImpl.class);
```
Ivan95kos, line 67: move this method to class field, initialize it there, not in this method
```suggestion
```

Ivan95kos (changes requested):

src/main/java/mate.academy/lib/Component.java
s-hrynus: @Ivan95kos What do you mean? 
Should I add this here @Target(ElementType.TYPE) or what?

aaananas (approved):

PR #679, opened 2023-07-10
Completed task

AntonU91:


src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Inject {
```
olekskov, line 7: Checklist item: "Do not forget to add @Target to custom annotations."

olekskov (approved):

src/main/java/mate.academy/lib/Inject.java
AntonU91: Done

PR #684, opened 2023-07-12
Homework jv-dependency-injection.

UniyatYura:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+        if (intances.containsKey(clazz)) {
+            return intances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            intances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                 | InstantiationException | IllegalAccessException e) {
```
okuzan, line 60: image

okuzan (changes requested):

PR #686, opened 2023-07-15
Add working solution: created new annotations - Component that indica…

WasjaZeikan:
…tes classes that the Injector can create and Inject that indicates fields that the Injector must initialize, class Injector that created and initialized objects; marked fields productParser and fileReaderService of class ProductServiceImpl with annotation Inject, classes FileReaderServiceImpl, ProductParserImpl, productServiceImpl with annotation Component


olekskov (approved):
Nice job

PR #691, opened 2023-07-24
Initial Solution

ArtemFedoseev:


kozhukhovsky (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,88 @@
...
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (!interfaceClazz.isInterface()) {
+            return interfaceClazz;
+        }
+        if (interfaceImplementations.size() == 0) {
+            interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+            interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
+            interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
+        }
```
Ivan95kos, line 85: For the future, it is better to set this data in the "interfaceImplementations" field than to check every time it is empty. you can initialize it with Map.of() method
this method creates an immutable map
``` suggestion
```

Ivan95kos (approved):
Great job, with minor comments

PR #698, opened 2023-08-02
implemented the custom dependency injection

ivan0dyatlyukkk:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,86 @@
...
+    private Map<Class<?>, Object> instances = new HashMap<>();
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Class<?> implementedClazz = findImplementation(interfaceClazz);
+        ifConsistComponentAnnotation(implementedClazz);
```
satori11111, line 29: Optional: I think consist is a bad word for this method, maybe you mean contains, or you can use present

satori11111 (approved):
Incredible solution!!!

Vasul-Pankiv (approved):
Great solution!

PR #699, opened 2023-08-03
Jv dependency injection solution

keda-vlad:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,139 @@
...
+        Map<Class<?>, Class<?>> interfacesImplementations = new HashMap<>();
+        for (Class<?> clazzInterface : findServiceInterfaces()) {
+            Class<?> interfaceImplementation = null;
+            for (Class<?> clazzImplementation : findComponents()) {
+                if (clazzInterface.isAssignableFrom(clazzImplementation)) {
+                    interfaceImplementation = clazzImplementation;
+                    break;
+                }
+            }
+            System.out.println(clazzInterface + " " + interfaceImplementation);
```
Denis-Balako, line 91: Why you are printing results?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,139 @@
...
+        String path = packageName.replace('.', '/');
+        ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
+        File packageDirectory = new File(classLoader.getResource(path).getFile());
+        File[] files = packageDirectory.listFiles();
+        return Arrays.stream(files)
+                .filter(File::isFile)
+                .filter(file -> file.getName().endsWith(CLASS_INDICATOR))
+                .map(file -> getClass(file.getName(), packageName))
+                .collect(Collectors.toList());
+    }
```
Denis-Balako, line 119: public methods should be before private

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,139 @@
...
+                throw new RuntimeException("Can`t find interface " + interfaceClazz.getName());
+            }
+            Class<?> implementation = interfacesImplementations.get(interfaceClazz);
+            if (implementation == null) {
+                throw new RuntimeException("Interface " + interfaceClazz.getName()
+                        + " has no implementations or they doesn't annotated by @Component");
+            }
+            return implementation;
+        }
+        componentCheck(interfaceClazz);
```
Denis-Balako, line 77: You should decide how to check exceptions: in separate methods or in this method.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,139 @@
...
+import java.lang.reflect.Field;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private static final Map<Class<?>, Class<?>> interfacesImplementations;
```
Denis-Balako, line 14: From my point of view, we need to follow KISS principles and use just "Map.of(...)" for interface implementations.

Denis-Balako (changes requested):
Good solution! Let's try to improve it

src/main/java/mate.academy/lib/Injector.java
keda-vlad: used for quick check, forgot to delete)

Denis-Balako (approved):
Great job!

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,92 @@
...
+        return interfaceClazz;
+    }
+
+    private static Map<Class<?>, Class<?>> interfacesImplementationsConfig() {
+        Map<Class<?>, Class<?>> map = new HashMap<>();
+        map.put(FileReaderService.class, FileReaderServiceImpl.class);
+        map.put(ProductParser.class, ProductParserImpl.class);
+        map.put(ProductService.class, ProductServiceImpl.class);
+        return map;
+    }
```
Vasul-Pankiv, line 77: Common mistakes: "Make Interface Implementations map a class field. You can fill it in using Map.of()"

Vasul-Pankiv (changes requested):
Greate solution!
But need some improvments

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,83 @@
...
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            INSTANCES.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create new instance of class - " + clazz.getName());
+        }
+    }
+
+    private static Class<?> findImplementation(Class<?> interfaceClazz) {
```
andrii-hoienko, line 61: Why all private methods is static?

andrii-hoienko (changes requested):
Minor fix required

nacenik (approved):
LGTM! 👍

PR #701, opened 2023-08-04
Made dependency injection for project logic. Added implementation for…

Denis-Balako:
… Component and Inject annotations, Injector.

### DIFF
--- a/src/main/java/mate.academy/lib/Component.java
+++ b/src/main/java/mate.academy/lib/Component.java
@@ -8,5 +8,4 @@ import java.lang.annotation.Target;
 @Retention(RetentionPolicy.RUNTIME)
 @Target(value = {ElementType.TYPE})
 public @interface Component {
-
 }
--- a/src/main/java/mate.academy/lib/Component.java
+++ b/src/main/java/mate.academy/lib/Component.java
@@ -1,5 +1,12 @@
 package mate.academy.lib;
 
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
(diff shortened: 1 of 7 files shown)


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+        }
+        return interfaceClazz;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Injection failed, "
+                    + "missing @Component annotation on the class "
+                    + clazz.getName());
+        }
```
holsoni, line 63: I think that you have to check if Class has Component Annotation not when creating an instance of class, but in the very beginning of your getInstance() method. Why do you need to do all of this operations, if the class doesn't  has this annotation?

holsoni (changes requested):
Nice solution, one small comment

holsoni (approved):

src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,11 @@
 package mate.academy.lib;
 
-public @interface Component {
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
 
+@Retention(RetentionPolicy.RUNTIME)
+@Target(value = {ElementType.TYPE})
```
Vasul-Pankiv, line 9:
```suggestion
@Target(ElementType.TYPE)
```

Vasul-Pankiv (approved):

PR #703, opened 2023-08-06
hw-dependency-injection

AntonZhdanov:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,85 @@
...
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Injection failed, "
+                    + "missing @Component annotation on the class: " + interfaceClazz);
```
OlegStashkiv, line 33: I think it is better to specify here:
`        throw new ComponentException("Injection failed, "
                    + "missing @Component annotation on the class: "
                    + clazz.getName());`
Because you check whether the annotation is in the implementation class

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,85 @@
...
+                }
+            }
+        }
+        if (clazzImplInstance == null) {
+            clazzImplInstance = createNewInstance(clazz);
+        }
+        return clazzImplInstance;
+    }
+
+    private Object createNewInstance(Class<?> clazz) throws RuntimeException {
```
OlegStashkiv, line 60: Why does it throw RuntimeException here?

OlegStashkiv (changes requested):
Good decision. But I think there are several points that should be corrected.

OlegStashkiv (approved):

vovasalyha (approved):

PR #704, opened 2023-08-07
Implemented Injector uses @Component and @Inject. Add customize excep…

OlegStashkiv:
…tion ComponentException.

### DIFF
--- a/src/main/java/mate.academy/lib/Component.java
+++ b/src/main/java/mate.academy/lib/Component.java
@@ -1,9 +1,12 @@
 package mate.academy.lib;
 
+import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
 
 @Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
 public @interface Component {
 
 }
--- a/src/main/java/mate.academy/lib/Component.java
+++ b/src/main/java/mate.academy/lib/Component.java
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Component {
 
 }
--- a/src/main/java/mate.academy/lib/Inject.java
+++ b/src/main/java/mate.academy/lib/Inject.java
@@ -1,9 +1,12 @@
 package mate.academy.lib;
 
+import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
 
 @Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
 public @interface Inject {
 
 }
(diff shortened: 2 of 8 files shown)


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Component {
```
Dimagaa, line 7: Add @Target annotation with an appropriate ElementType.

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Inject {
```
Dimagaa, line 7: Add @target annotation with an appropriate ElementType.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private static final Map<Class<?>, Class<?>> interfaceImplementations =
+            Map.of(FileReaderService.class, FileReaderServiceImpl.class,
+                    ProductParser.class, ProductParserImpl.class,
+                    ProductService.class, ProductServiceImpl.class);
+    private static final Map<Class<?>, Object> instances = new HashMap<>();
```
Dimagaa, line 21: I would recommend you use lazy initialization

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementationInstance = null;
```
Dimagaa, line 28: Let’s check instances map before new instance creation

Dimagaa: I mean that if there is such an instance in the 'instances' map, we don't need to iterate through the declared fields

Dimagaa (changes requested):

src/main/java/mate.academy/lib/Injector.java
OlegStashkiv: I understand, but I think leaving it as it will not be a mistake.

Dimagaa (approved):
Great job

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,79 @@
...
+        Class<?> clazz = findImplementation(interfaceClazz);
+        if (!clazz.isAnnotationPresent(Component.class)) {
```
vovasalyha, line 34: Not a big deal, but you can make your method less complex by extracting this check to a separate private method.

vovasalyha (approved):

PR #705, opened 2023-08-07
Implement Injector

Dimagaa:
Update Component and Injector annotation Retention policy and set Target element type

### DIFF
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,13 +1,76 @@
 package mate.academy.lib;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.util.HashMap;
+import java.util.Map;
+import mate.academy.service.FileReaderService;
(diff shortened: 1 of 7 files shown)


OlegTuryk (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+        return instances.getOrDefault(implementationClazz, createNewInstance(implementationClazz));
+    }
+
+    private Class<?> getComponentClass(Class<?> interfaceClazz) {
+        Class<?> componentClass = interfaceClazz.isInterface()
+                ? implementations.get(interfaceClazz) : interfaceClazz;
+        if (!componentClass.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Can't get instance of non-component class "
+                    + interfaceClazz.getName());
+        }
```
Vasul-Pankiv, line 44: move it to separate method

Vasul-Pankiv (changes requested):

PR #707, opened 2023-08-07
implemented injector

IvanFadieiev:


src/main/java/mate.academy/service/FileReaderService.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 
+@Component
```
keda-vlad, line 6: U don't need annotation here, only on implementation
"Convention is to use @Component on the classes that are used for instance creation (usually these are interface implementations) "

src/main/java/mate.academy/service/ProductParser.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
 public interface ProductParser {
```
keda-vlad, line 7: same

src/main/java/mate.academy/service/ProductService.java
```diff
@@ -1,8 +1,10 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
 public interface ProductService {
```
keda-vlad, line 8: same

keda-vlad (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        if (clazz.isAnnotationPresent(Component.class)) {
```
vovasalyha, line 29: It's better to check in if statement if annotation is absent and throw exception if true (better even create a separate method to check it). It'll make your code cleaner.

vovasalyha (approved):
Please see the comment.

keda-vlad (approved):

PR #709, opened 2023-08-08
implemented injector class

DmytroV95:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,79 @@
...
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
 
+    private final Map<Class<?>, Object> instances = new HashMap<>();
+    private final Map<Class<?>, Class<?>> interfaceImplementations = Map.of(
```
IhorKalaur, line 18: In common mistakes was a requirement
 `Make Interface Implementations map a class field.`
I think, we need set` static` type for this map.

IhorKalaur (approved):
Good, some detail need to fix

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,79 @@
...
+import java.util.Map;
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
```
vovasalyha, line 16:
```suggestion
```

vovasalyha (approved):

PR #710, opened 2023-08-08
Solution

ArtEg0:
Added all necessary annotations: @Component and @Inject.
Added logic to Injector class.
MVN clean package done.
All tests passed.
Code executing as intended.

### DIFF
--- a/src/main/java/mate.academy/lib/Inject.java
+++ b/src/main/java/mate.academy/lib/Inject.java
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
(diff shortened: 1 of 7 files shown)


src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Inject {
```
olekskov, line 7: Do not forget about Target annotation

olekskov (approved):
Good job, approved

PR #712, opened 2023-08-08
Create injector functions

ConfusedSatlan:


src/main/java/mate.academy/service/ProductParser.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
```
Someboty, line 6: Same here

src/main/java/mate.academy/service/ProductService.java
```diff
@@ -1,8 +1,10 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
```
Someboty, line 7: And here

src/main/java/mate.academy/service/FileReaderService.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 
+@Component
```
Someboty, line 6: Why do you need @Component both on the interface and its implementation?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,85 @@
...
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (InstantiationException
+                 | InvocationTargetException
+                 | NoSuchMethodException
+                 | IllegalAccessException e) {
```
Someboty, line 76: It is better to replace many exceptions that have a common parent with a general parental exception.

Bad example: 
} catch (InvocationTargetException | InstantiationException
      | IllegalAccessException | NoSuchMethodException e) {

Good example: 

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,85 @@
...
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
-    private static final Injector injector = new Injector();
+    private static final Map<Class<?>, Class<?>> interfacesImplementation;
+    private static final Injector injector;
+    private Map<Class<?>, Object> instances;
+
+    public Injector() {
+        instances = new HashMap<>();
```
Someboty, line 21: I don't sure you need it in the constructor

Someboty (changes requested):
Great work! Let's finish it)

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object newInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Class must be annotated with @Component");
+        }
```
Someboty, line 37: You can add it to the findImplementation() method or create a new one.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+    private static final Injector injector;
+    private Map<Class<?>, Object> instances = new HashMap<>();
+
+    static {
+        interfacesImplementation = Map.of(
+                FileReaderService.class, FileReaderServiceImpl.class,
+                ProductParser.class, ProductParserImpl.class,
+                ProductService.class, ProductServiceImpl.class
+        );
+        injector = new Injector();
```
vovasalyha, line 25: Initialize it immediately during declaration, it's not necessary to do it here, same with map of implementations.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object newInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Class must be annotated with @Component");
```
vovasalyha, line 36: The message is not informative, specify which class is not annotated. Also it's better to extract this check to separate method to make this method less complex.

vovasalyha (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance" + clazz.getName());
```
Someboty, line 69: When you throw an exception add an informative message to it. Also, don’t forget to add an exception you’re catching in the catch block to the RuntimeException object you throw

Bad example: 
} catch(Exception e) {
throw new SomeException("Can't initialize object with Injector");
}
Good example: 
} catch(Exception e) {
throw new SomeException("Injection failed, missing @Component annotaion on the class " + someInfoAboutClass, e);
}

src/main/java/mate.academy/lib/Injector.java
Someboty, line 25: Do you really need this block? You can use Map.of() and create a new Injector directly in the fields

Someboty (changes requested):
Great! Added some more suggestions to improve your solution)

nacenik (approved):
LGTM! 👍

PR #714, opened 2023-08-08
done solution

DaniilNepelev:


ArtemReizer (approved):
Good solution!

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Class<?> implementationClazz = getComponentClass(interfaceClazz);
+        return instances.getOrDefault(implementationClazz, createNewInstance(implementationClazz));
+    }
+
+    private Class<?> getComponentClass(Class<?> interfaceClazz) {
+        Class<?> componentClass = interfaceClazz.isInterface()
```
andrii-hoienko, line 32: Bad variable name it is not componentClass at least you dont know about that on this moment

andrii-hoienko (changes requested):
Nice, minor fix and you are done

nacenik (approved):
LGTM! 👍

PR #715, opened 2023-08-08
dependecy-injection-injector-hw

ihorvasylenko23:


nacenik (changes requested):
Good solution, let's improve it)

nacenik (approved):
LGTM! 👍

PR #718, opened 2023-08-08
implemented task

OleksandrSaven:


src/main/java/mate.academy/Main.java
```diff
@@ -10,7 +10,8 @@ public class Main {
     public static void main(String[] args) {
         // Please test your Injector here. Feel free to push this class as a part of your solution
         Injector injector = Injector.getInjector();
-        ProductService productService = null;
+        Object instance = injector.getInstance(ProductService.class);
+        ProductService productService = (ProductService) instance;
```
SaiyaJ1N, line 14: You can cast to ProductService in one line
```suggestion
        ProductService productService = (ProductService) injector.getInstance(ProductServiceImpl.class);
```

SaiyaJ1N (approved):
Your solution is great!

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,68 @@
...
+                try {
+                    field.setAccessible(true);
+                    field.set(clazzImplementationIns, fieldInstance);
+                } catch (IllegalAccessException e) {
+                    throw new RuntimeException("Can't initialize field."
+                            + " Class " + clazz.getName() + "field " + field.getName());
+                }
+            }
+        }
+        if (!clazz.isAnnotationPresent(Component.class)) {
```
andrii-hoienko, line 42: so first you will execute some difficult logic and only in the end you will check hmmm maybe it is not even componen so evetything I done defore is not needed right?

andrii-hoienko (changes requested):
Nice, but some improvements required

nacenik (approved):
LGTM! 👍

PR #722, opened 2023-08-08
added solution

Visqe:


irynamekh (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (!interfaceClazz.isInterface()) {
+            return interfaceClazz;
+        }
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
+        return interfaceImplementations.get(interfaceClazz);
```
Vasul-Pankiv, line 54: make map a class field and initialize it using Map.of()

Vasul-Pankiv (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -14,6 +14,11 @@
 public class Injector {
     private static final Injector injector = new Injector();
     private final Map<Class<?>, Object> instances = new HashMap<>();
+    private final Map<Class<?>, Class<?>> interfaceImplementations = Map.of(
```
vovasalyha, line 17: Class field means than it should be static.

vovasalyha (approved):

PR #723, opened 2023-08-08
Implemented Field Injection

Zel1oy:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,71 @@
...
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object classImplementationInstance = null;
+        Class<?> classImpl = findImplementationClass(interfaceClazz);
+        if (!classImpl.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Can't initialize object without Component annotation");
```
nacenik, line 31: Which object doesn't have this annotation?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,71 @@
...
+        for (Field field : declaredFields) {
+            if (field.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = getInstance(field.getType());
+                classImplementationInstance = createNewInstance(classImpl);
+                try {
+                    field.setAccessible(true);
+                    field.set(classImplementationInstance, fieldInstance);
+                } catch (IllegalAccessException e) {
+                    throw new RuntimeException("Can't initialize field value. "
+                            + "Class: " + classImpl.getName() + "Field: " + field.getName());
```
nacenik, line 43: don't forget to put a stack trace from the exception 
```suggestion
                    throw new RuntimeException("Can't initialize field value. "
                            + "Class: " + classImpl.getName() + "Field: " + field.getName(), e);
```

nacenik (changes requested):
Don't forget to test your solution in the main

src/main/java/mate.academy/lib/Injector.java
nacenik, line 31: Create additional method for this check

vovasalyha (approved):
Approving, but please take a look on my comments.

PR #735, opened 2023-08-29
Attempt-1

AnatolyNz:


kozhukhovsky (changes requested):
Java Cl checks failed. Before making a Pull Request (or a new commit to an already existing PR), you need to write the `mvn clean package` command in the terminal and make sure that all checks pass. After that, the mentor will be able to review the homework

src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Component {
```
kozhukhovsky, line 7: add target annotation

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of "
+                    + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
```
kozhukhovsky, line 65: move this map to the class field and init, like:

```
);
```

kozhukhovsky (changes requested):

PR #739, opened 2023-09-06
created own Injector also Inject and Component annotation

DenysMaksymKachalup:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Component {
```

Rommelua (changes requested):

sokimaaa (approved):

PR #741, opened 2023-09-13
completed jv-dependency-injection task

Ka1zerov:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Component {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Component {
```
Ivan95kos, line 7: add target annotation

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Inject {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
```
Ivan95kos, line 6: the same

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can`t create an instance of " + clazz.getName());
+        }
+
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplMap = new HashMap<>();
```
Ivan95kos, line 65: move this map to the class field and init, like:

```
);
```

Ivan95kos (changes requested):

olekskov (approved):

PR #742, opened 2023-09-13
Dependency-injection impl

Oleksandr-Tymoshenko:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Component {
```
NazarSmith228, line 7: it is a good practice to always specify `@Target`

src/main/java/mate.academy/service/ProductService.java
```diff
@@ -1,8 +1,10 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
```
NazarSmith228, line 7: interface should not contain this annotation, since you can't instantiate it

src/main/java/mate.academy/service/ProductParser.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
```
NazarSmith228, line 6: interface should not contain this annotation, since you can't instantiate it

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+            throw new RuntimeException("Injection failed, missing @Component"
+                    + " annotation on the class " + interfaceClazz.getSimpleName());
+        }
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        Field[] declaredFields = clazz.getDeclaredFields();
+        for (var field : declaredFields) {
+            if (field.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = getInstance(field.getType());
+                clazzImplementationInstance = createNewInstance(clazz);
```
NazarSmith228, line 37: what is the point of re-creating the instance in a loop?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+            ProductService.class, ProductServiceImpl.class);
 
     public static Injector getInjector() {
         return injector;
     }
 
-    public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+    public Object getInstance(@NotNull Class<?> interfaceClazz) {
+        if (!interfaceClazz.isAnnotationPresent(Component.class)) {
```
NazarSmith228, line 27: you should not check the annotation on the interface

instead you should find an available implementation of that interface and verify annotation there

this can be done in a method `findComponentAnnotatedImplementation`

NazarSmith228 (changes requested):
Good job! Let's improve your solution

liliia-ponomarenko (approved):
Good job ;)

PR #744, opened 2023-09-13
implemented dependency injection

RedArtelerist:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,82 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object newInstance = constructor.newInstance();
+            instances.put(clazz, newInstance);
+            return newInstance;
+        } catch (NoSuchMethodException | InstantiationException
+                 | IllegalAccessException | InvocationTargetException e) {
```
liliia-ponomarenko, line 72:
```suggestion
        } catch (ReflectiveOperationException e) {
```

liliia-ponomarenko (changes requested):

liliia-ponomarenko (approved):
Great! 👍

PR #745, opened 2023-09-14
Realized injector class for annotation Inject and Component.

Vlad1507:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create an instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> componentMap = new HashMap<>();
```
sokimaaa, line 65: lets make this map class level field and initialize it accordingly: 

```
);
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+        componentMap.put(ProductParser.class, ProductParserImpl.class);
+        componentMap.put(ProductService.class, ProductServiceImpl.class);
+        componentMap.put(FileReaderService.class, FileReaderServiceImpl.class);
+        if (componentMap.containsKey(interfaceClazz)) {
```
sokimaaa, line 69: add check `interfaceClazz` is interface, and in this case should get implementation from componentMap

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+                }
+            }
+        } else {
+            throw new RuntimeException("Injection failed, missing @Component "
+                    + "annotation on the class " + clazz.getName());
+        }
+        if (clazzImplementationInstance == null) {
+            clazzImplementationInstance = createNewInstance(clazz);
+        }
+        return clazzImplementationInstance;
```
sokimaaa, line 47: Revert if-else statement to reduce code, e.g. 
```
if (!iclazz.isAnnotationPresent(Component.class)) {
/** throwing exception **/
}

/** code in case annotation is present **/
```

sokimaaa (changes requested):

sokimaaa (approved):

PR #747, opened 2023-09-15
Implemented Injector class, Component, Inject.

ystankevych:


teract10s (approved):
good job

liliia-ponomarenko (changes requested):
Let’s  initialize `ProductService productService` in Main

PR #749, opened 2023-09-17
implemented Injector

dmytrocherepov:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,78 @@
...
+    private Object createNewInstance(Class<?> clazz) {
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        Object instance;
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
```
ystankevych, line 69:
```suggestion
        try {
            Constructor<?> constructor = clazz.getConstructor();
           Object instance = constructor.newInstance();
            instances.put(clazz, instance);
            return instance;
```
You return you instance from try block, so you don't need to declare it outside try block.

ystankevych (approved):
Hi, Dmytro. Great job! I approved but left one comment =)

liliia-ponomarenko (approved):
LGTM 🔥

PR #750, opened 2023-09-17
implemented injector

mmosii:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
dmytrocherepov, line 6: From the 'Common mistakes" Do not forget to add https://github.com/target to custom annotations.

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
```
dmytrocherepov, line 5: same here

dmytrocherepov (changes requested):

dmytrocherepov (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
```
liliia-ponomarenko, line 65: It should be a class-level variable to avoid creating a Map every time the method getInstance calls.

liliia-ponomarenko (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
```
liliia-ponomarenko, line 68: From the previous review: Also, you can initialize it in static{} block using Map.of()
Now every time you call getInstance you will put these implementations. Is it really needed?

liliia-ponomarenko (changes requested):

src/main/java/mate.academy/lib/Injector.java
mmosii: Sorry, my bad, didn't get it before)

liliia-ponomarenko (approved):
great 😊

PR #753, opened 2023-09-18
'Dependency-injection' task is implemented

Alexsandr-Yablunovskyi:


don-bigdad (approved):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+            clazzImplInstance = createNewInstance(clazz);
+        }
+        return clazzImplInstance;
+    }
+
+    private Class<?> findImpl(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImpl = new HashMap<>();
+        interfaceImpl.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImpl.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImpl.put(ProductService.class, ProductServiceImpl.class);
```
liliia-ponomarenko, line 55: It should be a class-level variable to avoid creating a Map every time the method getInstance calls.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                 | InstantiationException | IllegalAccessException e) {
```
liliia-ponomarenko, line 72:
```suggestion
        } catch (ReflectiveOperationException e) {
```

liliia-ponomarenko (changes requested):
Good job! Let’s fix a few mistakes  ;)

Elena-Bruyako (approved):

PR #757, opened 2023-09-19
add component annotation and modofy injector

yuriy49:


src/main/java/mate.academy/service/FileReaderService.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 
+@Component
```
olekskov, line 6: this annotation is required on implementations, not interfaces

olekskov (approved):
Approved but please pay attention to comment below for further

PR #760, opened 2023-09-19
implemented injector and component

TetianaDenchyk:


src/main/java/mate.academy/Main.java
```diff
@@ -10,7 +10,7 @@ public class Main {
     public static void main(String[] args) {
         // Please test your Injector here. Feel free to push this class as a part of your solution
```
validatorr, line 11: redundant line

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+        }
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException | InstantiationException
```
validatorr, line 60: You can replace all of these exceptions only by one general parental exception

validatorr (approved):
LGTM! Some minor fixes to make your code cleaner

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of  " + clazz.getName());
+        }
+    }
+
+    private static Class<?> getImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+    private Map<Class<?>, Object> instances = new HashMap<>();
+
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = getImplementation(interfaceClazz);
```
Elena-Bruyako, line 25: You need first to check if annotation Component present 
            throw new RuntimeException( "Annotation Component is missing. "
                    + "Can't create instance of a class " + clazz.getName());
        }

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+import java.util.Map;
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
```
Elena-Bruyako, line 16: Remove empty line

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #761, opened 2023-09-19
implemented Injector + annotations

validatorr:


vsych321 (approved):
Good good!

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+            Object newInstance = constructor.newInstance();
+            instances.put(clazz, newInstance);
+            return newInstance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create an instance of " + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
```

src/main/java/mate.academy/service/FileReaderService.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 
+@Component
```
Elena-Bruyako, line 6: Annotation Component is redundant here

src/main/java/mate.academy/service/ProductParser.java
```diff
@@ -1,12 +1,15 @@
 package mate.academy.service;
 
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
```
Elena-Bruyako, line 6: The same

src/main/java/mate.academy/service/ProductService.java
```diff
@@ -1,8 +1,10 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
```
Elena-Bruyako, line 7: The same

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #763, opened 2023-09-19
Implemented all the methods and classes

Cayman4010:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,81 @@
...
+    private static final Map<Class<?>, Object> instances = new HashMap<>();
+    private static final Map<Class<?>, Class<?>> interfaceImplementations;
+
+    static {
+        interfaceImplementations = Map.of(
+                FileReaderService.class, FileReaderServiceImpl.class,
+                ProductParser.class, ProductParserImpl.class,
+                ProductService.class, ProductServiceImpl.class
+        );
+    }
```
torbinofficial, line 25: You can initialize instances map as local variable in findImplementation, but that's up to u

torbinofficial (commented):
Great job!
1 minor comment, but nothing to change, imho

torbinofficial (approved):

src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Component {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
```

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Inject {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
```

src/main/java/mate.academy/lib/Injector.java
Elena-Bruyako: Not agree. In such way the Map will be created every time when the method getInstance calls.

Elena-Bruyako (changes requested):
Good job!

liliia-ponomarenko (approved):
Great 😊

PR #768, opened 2023-09-20
Added solution

AlexVulpine:
Implemented Injector class and added annotations

### DIFF
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -14,7 +14,6 @@ import mate.academy.service.impl.ProductServiceImpl;
 public class Injector {
     private static final Injector injector = new Injector();
     private static final Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
-    private final Map<Class<?>, Object> instances = new HashMap<>();
 
     static {
         interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
@@ -35,19 +34,15 @@ public class Injector {
     }
 
     private Object createOrGetInstance(Class<?> interfaceClazz) {
-        if (instances.containsKey(interfaceClazz)) {
-            return instances.get(interfaceClazz);
+        Class<?> implementationClass = interfaceImplementations.getOrDefault(interfaceClazz,
+                                                                            interfaceClazz);
+        if (implementationClass.isAnnotationPresent(Component.class)) {
+            Object instance = createNewInstance(implementationClass);
+            injectDependencies(instance);
+            return instance;
         }
-        Class<?> implementationClass = interfaceImplementations
-                .getOrDefault(interfaceClazz, interfaceClazz);
-        if (!implementationClass.isAnnotationPresent(Component.class)) {
-            throw new RuntimeException(Component.class + " annotation missing. "
-                    + "Can't create an instance of class " + implementationClass.getName());
-        }
-        Object instance = createNewInstance(implementationClass);
-        injectDependencies(instance);
-        instances.put(interfaceClazz, instance);
-        return instance;
+        throw new RuntimeException(Component.class + " annotation missing. "
+                + "Can't create an instance of class " + implementationClass.getName());
     }
 
     private Object createNewInstance(Class<?> clazz) {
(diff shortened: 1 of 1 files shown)


dmytrocherepov (approved):
Goog Job!

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,82 @@
...
-    private static final Injector injector = new Injector();
+    private static final Injector INJECTOR = new Injector();
+    private final Map<Class<?>, Object> instances = new HashMap<>();
+    private final Map<Class<?>, Class<?>> interfaceImplementations = Map.of(
+            FileReaderService.class, FileReaderServiceImpl.class,
+            ProductService.class, ProductServiceImpl.class,
+            ProductParser.class, ProductParserImpl.class
+    );
+
+    private Injector() {
```
Elena-Bruyako, line 23: What is the purpose of this?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,82 @@
...
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
-    private static final Injector injector = new Injector();
+    private static final Injector INJECTOR = new Injector();
+    private final Map<Class<?>, Object> instances = new HashMap<>();
+    private final Map<Class<?>, Class<?>> interfaceImplementations = Map.of(
```
Elena-Bruyako, line 17: Initialise the Map in static block

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,82 @@
...
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
-    private static final Injector injector = new Injector();
+    private static final Injector INJECTOR = new Injector();
```
Elena-Bruyako, line 15: It's redundant changing

src/main/java/mate.academy/service/impl/FileReaderServiceImpl.java
```diff
@@ -4,9 +4,15 @@
 import java.io.IOException;
 import java.nio.file.Files;
 import java.util.List;
+import mate.academy.lib.Component;
 import mate.academy.service.FileReaderService;
 
+@Component
 public class FileReaderServiceImpl implements FileReaderService {
+    public FileReaderServiceImpl() {
```
Elena-Bruyako, line 12: Remove

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,82 @@
...
+
+        if (!implementationClass.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException(Component.class + " annotation missing. "
+                    + "Can't create an instance of class " + implementationClass.getName());
+        }
+
+        Object instance = createNewInstance(implementationClass);
+
+        injectDependencies(instance);
+
```
Elena-Bruyako, line 50: A lot of empty lines in this method

Elena-Bruyako (changes requested):

src/main/java/mate.academy/lib/Injector.java
AlexRosenkreuz: I updated default constructor parameter to private, so that it was impossible to create an instance of injector class outside of injector itself. It forces to use only getInjector method

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,80 @@
...
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private static final Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
+    private final Map<Class<?>, Object> instances = new HashMap<>();
```
NazarSmith228, line 17: Could you explain to me the necessity of this cache? I don't see such requirement.

Neither do I say it is incorrect. I'm just trying to undestand the reason. We have some injector, that can create class instances and inject their dependencies via Reflection API. But internally it also has some instance cache - which is unknown to the user. So what is the motivation for such implementation detail?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,80 @@
...
+            return constructor.newInstance();
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName(), e);
+        }
+    }
+
+    private void injectDependencies(Object instance) {
+        Field[] declaredFields = instance.getClass().getDeclaredFields();
+        for (Field field : declaredFields) {
+            if (field.isAnnotationPresent(Inject.class)) {
```
NazarSmith228, line 66: the content of this `if` can be extracted into a separate method, which will make this method simpler

NazarSmith228 (changes requested):
Great job! Let's improve your solution

src/main/java/mate.academy/lib/Injector.java
AlexRosenkreuz: I revised the purpose. Originally it should have boosted performance as it returns the object from the map, if it has already been created. But it seems to ruin the Immutability so I got rid of it.

Andriy-dz (approved):

PR #776, opened 2023-09-22
implemented Injector.java. and added @Component annotations to classe…

opopovych:
…sImpl.

### DIFF
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,13 +1,74 @@
 package mate.academy.lib;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.util.HashMap;
+import java.util.Map;
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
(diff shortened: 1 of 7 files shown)


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>,Class<?>> interfaceImplementations = new HashMap<>();
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
+        if (interfaceClazz.isInterface()) {
+            return interfaceImplementations.get(interfaceClazz);
+        }
+        return interfaceClazz;
```
sokimaaa, line 72: map with accordant beans should be a class field 

It can be initialized this way, e.g. 
```
private static final Map<...> implementations = Map.of(...);
```

or initialized through `static constructor`

sokimaaa (changes requested):

sokimaaa (approved):

PR #784, opened 2023-10-06
Hw dependency injection

kaptetos:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
okuzan, line 6: image

okuzan (changes requested):
Hey there! 😊 Before you submit your work for review, please go through the checklist under the task and make any necessary corrections. 📝✅ Your mentor will be happy to help once those points are sorted out.  Don't worry, it's all part of the learning process! 📚✨

src/main/java/mate.academy/lib/Component.java
kaptetos: I'm sorry, I'll fix it. I'll be more careful

Ivan95kos (approved):
Nice

PR #794, opened 2023-10-21
Roma Menzheres's realization of 'Java Dependency Injection' task

RomanMenzheres:
Created '@Inject' annotation to mark field which will be injected, '@Component' annotaion to mark classes instance of which will be created. Implemented 'getInstance' method which return instance of class with all needed dependecies

### DIFF
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,13 +1,87 @@
 package mate.academy.lib;
 
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.util.HashMap;
+import java.util.Map;
+import mate.academy.service.FileReaderService;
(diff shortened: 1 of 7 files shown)


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,87 @@
...
+
+        try {
+            Object instance = implementationClazz.getConstructor().newInstance();
+            instances.put(implementationClazz, instance);
+            return instance;
+        } catch (InstantiationException | IllegalAccessException
+                 | InvocationTargetException | NoSuchMethodException e) {
+            throw new RuntimeException("Can't create an instance of "
+                    + implementationClazz + " class", e);
+        }
```
Ivan95kos, line 77: it is better to use one `ReflectiveOperationException`
or add the `catch` for each Exception

Ivan95kos (approved):
Great job, with a minor comment

PR #797, opened 2023-10-22
implemented the injector

flekros:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private Map<Class<?>, Class<?>> implementaionMap = new HashMap<>();
+    private Map<Class<?>, Object> instances = new HashMap<>();
```
kshuryhin, line 17: Probably we can also make them `final`

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+        }
+        if (clazzImplementatiomInstance == null) {
+            clazzImplementatiomInstance = createNewInstance(implementationClazz);
+        }
+        return clazzImplementatiomInstance;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
+        if (instances.containsKey(clazz)) {
+            return implementaionMap.get(clazz);
```
kshuryhin, line 53: Are you sure we should get value from `implementationMap` here?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object clazzInstance = constructor.newInstance();
+            instances.put(clazz.getClass(), clazzInstance);
```
kshuryhin, line 58: `clazz.getClass` is something do not we want to achieve. See previous comment regarding instances check and try to update your code

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+        if (!implementationClazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Injection failed,"
+                    + " missing @Component annotation on the class "
+                    + implementationClazz.getName());
+        }
+        Field [] declaredField = implementationClazz.getDeclaredFields();
+        for (Field field : declaredField) {
+            if (field.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = getInstance(field.getType());
+                clazzImplementatiomInstance = createNewInstance(implementationClazz);
```
kshuryhin, line 35: I believe we do not want our instance creation inside the loop, what do you think?

kshuryhin (changes requested):
Good job, but there are some points to consider

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,70 @@
...
+            return clazzInstance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        implementaionMap.put(FileReaderService.class, FileReaderServiceImpl.class);
+        implementaionMap.put(ProductParser.class, ProductParserImpl.class);
+        implementaionMap.put(ProductService.class, ProductServiceImpl.class);
```
bhdnchui, line 64: Looks like map initialization. I think it would be better not to make it in this method.

bhdnchui (changes requested):
Good job!

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private static final Map<Class<?>, Class<?>> implementaionMap = new HashMap<>();
+    private final Map<Class<?>, Object> instances = new HashMap<>();
 
     public static Injector getInjector() {
+        initializationMap(implementaionMap);
```
bhdnchui, line 20: Use a verb in method naming. Also it is no need to pass a map to method. 
One more point. If you map is private static final, it is constant and it's content can be initialized as constant.
```suggestion
        initializeMap();
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private static final Map<Class<?>, Class<?>> implementaionMap = new HashMap<>();
```
bhdnchui, line 16: Initialize map with `Map.of()` syntax.

bhdnchui (changes requested):
A few points to consider.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,71 @@
...
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private static final Map<Class<?>, Class<?>> implementaionMap
+            = Map.of(FileReaderService.class, FileReaderServiceImpl.class,
```
bhdnchui, line 17: Make better code formatting here. Do not start a new line with =.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,71 @@
...
+        }
     }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (interfaceClazz.isInterface()) {
+            return implementaionMap.get(interfaceClazz);
+        }
+        return interfaceClazz;
+    }
+
```
bhdnchui, line 70:
```suggestion
```

bhdnchui (changes requested):
A few minor points.

bhdnchui (approved):
Good job

PR #808, opened 2023-11-06
realized Injector class

fedorovychh:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,87 @@
...
+                    throw new RuntimeException(String
+                            .format(FIELD_SET_FAILED_MESSAGE,
+                                    clazz.getName(), field.getName()), ex);
+                }
+            }
+        }
+        return clazzImplInstance == null
+                ? createNewInstance(clazz)
+                : clazzImplInstance;
+    }
```
YevheniiVlasenko, line 60: You can split it into 3 methods. At least into two by getting the check out from the method body

YevheniiVlasenko (changes requested):
Good job! Let's make some improvements for better readability!

bhdnchui (approved):
Good job!

PR #809, opened 2023-11-07
jv-dependency-injection solutin

YevheniiVlasenko:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,82 @@
...
+        for (Field field : declaredFields) {
+            if (field.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = getInstance(field.getType());
+                try {
+                    field.setAccessible(true);
+                    field.set(instance, fieldInstance);
+                } catch (IllegalAccessException e) {
+                    throw new RuntimeException("Can't initialize field value. "
+                        + "Class " + instance.getClass().getName()
+                        + ", field " + field.getName());
```
vadymhrnk, line 56: Don't forget to add an exception you're catching in catch block to the RuntimeException object you throw
```suggestion
                    throw new RuntimeException("Can't initialize field value. "
                        + "Class " + instance.getClass().getName()
                        + ", field " + field.getName(), e);
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,82 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
```
vadymhrnk, line 72: Same here
```suggestion
            throw new RuntimeException("Can't create a new instance of " + clazz.getName(), e);
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,82 @@
...
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
+    private static final HashMap<Class<?>, Class<?>> implementationMap = new HashMap<>();
+
+    static {
+        implementationMap.put(FileReaderService.class, FileReaderServiceImpl.class);
+        implementationMap.put(ProductParser.class, ProductParserImpl.class);
+        implementationMap.put(ProductService.class, ProductServiceImpl.class);
+    }
```
vadymhrnk, line 21: Only a suggestion, but this map could be initialized using `Map.of()`

vadymhrnk (changes requested):
Good job. Only few minor things to change

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,79 @@
...
+        for (Field field : declaredFields) {
+            if (field.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = getInstance(field.getType());
+                try {
+                    field.setAccessible(true);
+                    field.set(instance, fieldInstance);
+                } catch (IllegalAccessException e) {
+                    throw new RuntimeException("Can't initialize field value. "
+                        + "Class " + instance.getClass().getName()
+                        + ", field " + field.getName(), e);
```
bhdnchui, line 53: It would be better to use `String.format()`

bhdnchui (approved):
Good job!

PR #814, opened 2023-11-07
Solution for jv-dependency-injection task

AgroFix:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+                    throw new RuntimeException("Can not initial field value. "
+                            + "Class: " + clazz.getName() + ". Field: " + field.getName());
+                }
+            }
+        }
+        if (clazzImplementationInstance == null) {
+            clazzImplementationInstance = getOrCreateNewInstance(clazz);
+        }
+        return clazzImplementationInstance;
+    }
```
YevheniiVlasenko, line 49: I would split this method into two: getInstance() itself and injectFields()
But I don't think it's mandatory.

YevheniiVlasenko (changes requested):
Good job! Just a moment to make it perfect:

src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,12 @@
 package mate.academy.lib;
 
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
 public @interface Component {
```
d1sam, line 11:
```suggestion
```

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,12 @@
 package mate.academy.lib;
 
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
 public @interface Inject {
```
d1sam, line 11:
```suggestion
```

d1sam (approved):
Well done!

PR #816, opened 2023-11-07
implemented injector first version

annton-shakur:


src/main/java/mate.academy/service/FileReaderService.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 
+@Component
 public interface FileReaderService {
     List<String> readFile(String fileName);
 }
```
TarasYurkovskyi, line 9: Interface is not Component.  Remove it for all interfaces

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Inject {
 
 }
```
TarasYurkovskyi, line 9: remove empty line here and in Component too

TarasYurkovskyi (changes requested):
Good job but need to add some fixes

kshuryhin (approved):

PR #826, opened 2023-11-14
Hw solution injection

IhorTrokhymchuk:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Component {
```
Sanderfuck, line 7: What about @Target anno?

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Inject {
```
Sanderfuck, line 7: The same

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,83 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InstantiationException
+                  | IllegalAccessException | InvocationTargetException e) {
```
Sanderfuck, line 65: Can you catch one general exception for this all?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,83 @@
...
+        }
+    }
+
+    private Class<?> findImpl(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImpl = new HashMap<>();
+
+        interfaceImpl.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImpl.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImpl.put(ProductService.class, ProductServiceImpl.class);
+
```
Sanderfuck, line 76: for every invoke this method you will init map, don`t do this here

Sanderfuck (changes requested):

src/main/java/mate.academy/lib/Injector.java
IhorTrokhymchuk: in lesson use this catch block 
general exception for this it's Exception e?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,81 @@
...
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private static Map<Class<?>, Class<?>> interfaceImpl;
+    private Map<Class<?>, Object> instances = new HashMap<>();
```
okuzan, line 17: better use Map.of and initialize here, make it static

okuzan (approved):
Good job

src/main/java/mate.academy/lib/Injector.java
okuzan: interfaceImpl*

PR #829, opened 2023-11-21
implement injector

Paradoxallist:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,12 @@
 package mate.academy.lib;
 
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE_USE)
```
kerrrusha, line 9: i think you want to use this
```suggestion
@Target(ElementType.TYPE)
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>,Class<?>> interfaceImplementation = new HashMap<>();
+        interfaceImplementation.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementation.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementation.put(ProductService.class, ProductServiceImpl.class);
```
kerrrusha, line 71: you are going to create `interfaceImplementation` Map every time when we call this method? consider creating class-level `private final` field for Map

kerrrusha (changes requested):

Rommelua (approved):

PR #836, opened 2023-12-01
Implemented an injector with @Component and @Inject annotations to in…

misha-levko:
…itialize and return Object instances.

### DIFF
--- a/src/main/java/mate.academy/lib/Inject.java
+++ b/src/main/java/mate.academy/lib/Inject.java
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Inject {
+import java.lang.annotation.Retention;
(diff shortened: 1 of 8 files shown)


src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Inject {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
```
sokimaaa, line 6: use @Target for any annotation you create

Rommelua (changes requested):

kerrrusha (approved):

PR #837, opened 2023-12-04
first_commit

Lupen07:


src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,13 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.FIELD)
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Inject {
 
 }
+
```
kozhukhovsky, line 13:
```suggestion
```

kozhukhovsky (approved):
And not fixed previous comments.

PR #844, opened 2023-12-13
Injector implementation created using Component annotation

ivanlavrinenko:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
okuzan, line 6: add @Target
Make sure to avoid these mistakes
https://github.com/mate-academy/jv-dependency-injection/blob/main/checklist.md

okuzan (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,80 @@
...
+        } catch (InvocationTargetException | InstantiationException | IllegalAccessException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementationsMap = new HashMap<>();
+        interfaceImplementationsMap.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementationsMap.put(ProductService.class, ProductServiceImpl.class);
+        interfaceImplementationsMap.put(ProductParser.class, ProductParserImpl.class);
```
okuzan, line 76: Use map.of , make it a class field

kozhukhovsky (approved):

PR #845, opened 2023-12-15
It's injector implementation

VKhmil:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,85 @@
...
+            throw new RuntimeException(CANT_CREATE_AN_OBJECT + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (interfaceClazz.isInterface()) {
+            return interfaceImplementations.get(interfaceClazz);
+        } else {
+            throw new RuntimeException(NO_FOUND_IMPL + interfaceClazz.getName());
+        }
```
sokimaaa, line 83: u dont need else clause here
```suggestion
        if (interfaceClazz.isInterface()) {
            return interfaceImplementations.get(interfaceClazz);
        } 
        throw new RuntimeException(NO_FOUND_IMPL + interfaceClazz.getName());
```

sokimaaa (approved):

PR #858, opened 2023-12-30
solution 30/12

VadimKadasiev:


src/main/java/mate.academy/service/ProductService.java
```diff
@@ -1,8 +1,10 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
```
olekskov, line 7: we should place this annotation on implementations, not interfaces
```suggestion
```

olekskov (approved):
Good job overall, please pay attention to comment for future

src/main/java/mate.academy/service/ProductService.java
VadimKadasiev: OK. Thank you!

PR #859, opened 2024-01-02
Implemented an injector that uses the @Component and @Inject annotati…

mhorianin:
…ons to initialize and return Object instances.

### DIFF
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,13 +1,75 @@
 package mate.academy.lib;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.util.HashMap;
+import java.util.Map;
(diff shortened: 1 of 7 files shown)


okuzan (changes requested):
Please revisit common mistakes for this task

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
+        }
+
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations =
+                Map.of(FileReaderService.class, FileReaderServiceImpl.class,
+                        ProductParser.class, ProductParserImpl.class,
+                        ProductService.class, ProductServiceImpl.class);
```
okuzan, line 67: make it a private class field

okuzan (approved):

PR #872, opened 2024-01-30
done injector-hw

RomanSoproniuk:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementation
+                = Map.of(
+                        ProductService.class, ProductServiceImpl.class,
+                        ProductParser.class, ProductParserImpl.class,
+                        FileReaderService.class, FileReaderServiceImpl.class);
```
okuzan, line 70: init outside of the method, make it a class field

okuzan (changes requested):

kozhukhovsky (approved):

PR #876, opened 2024-02-03
implemented Injector

letmerelaxpls:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                 | InstantiationException | IllegalAccessException e) {
+            throw new RuntimeException("Can`t create a new instance of " + clazz.getName());
```
okuzan, line 61: checklist issue

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> implementations = new HashMap<>();
+        implementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        implementations.put(ProductParser.class, ProductParserImpl.class);
+        implementations.put(ProductService.class, ProductServiceImpl.class);
+        if (interfaceClazz.isInterface()) {
+            return implementations.get(interfaceClazz);
+        }
```
okuzan, line 72: no need to recreate this map per each method call, make it a class field

okuzan (changes requested):
Make sure to fix common mistakes mentioned in the checklist before re-requesting review

okuzan (changes requested):
ㅤ

Rommelua (approved):

PR #883, opened 2024-02-24
implementedIngectorAndComponent

shumYevh:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```

okuzan (changes requested):

okuzan (approved):

PR #886, opened 2024-03-01
injector implementation

orest40th:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,81 @@
...
+            throw new InjectionException("Cannot create a new instance for " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = Map.of(
+                FileReaderService.class, FileReaderServiceImpl.class,
+                ProductParser.class, ProductParserImpl.class,
+                ProductService.class, ProductServiceImpl.class
+        );
```
kerrrusha, line 74: this should be initialized once, not each time when method is being called
```suggestion
```

src/main/java/mate.academy/service/ProductParser.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
```
kerrrusha, line 6: this does nothing. we can't create instance of interface
```suggestion
```

src/main/java/mate.academy/service/FileReaderService.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 
+@Component
```
kerrrusha, line 6: this does nothing. we can't create instance of interface
```suggestion
```

src/main/java/mate.academy/service/ProductService.java
```diff
@@ -1,8 +1,10 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
```
kerrrusha, line 7: this does nothing. we can't create instance of interface
```suggestion
```

kerrrusha (changes requested):

Rommelua (approved):

PR #897, opened 2024-03-18
main version

EduarfF:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,57 @@
...
+                     | InstantiationException
+                     | InvocationTargetException e) {
+                throw new RuntimeException("Failed to instantiate class: " + clazz.getName(), e);
+            }
+        }
+        return instances.get(clazz);
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
```
makaroshyna, line 47: Chacklist:
"Make Interface Implementations map a class field. You can fill it in using Map.of()."

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,57 @@
...
+                Object instance = constructor.newInstance();
+                instances.put(clazz, instance);
+                for (Field field : clazz.getDeclaredFields()) {
+                    if (field.isAnnotationPresent(Inject.class)) {
+                        Object fieldInstance = getInstance(field.getType());
+                        field.setAccessible(true);
+                        field.set(instance, fieldInstance);
+                    }
+                }
+            } catch (NoSuchMethodException | IllegalAccessException
```
makaroshyna, line 37: Checklist:
image

src/main/java/mate.academy/lib/Injector.java
makaroshyna, line 1: You don't have a check if Component annotation is missing above the class:
"If the Component annotation is missing above the class, we shouldn't be able to create an instance of this class in Injector, and we should throw an exception"

makaroshyna (changes requested):
Good! Check requirements in checklist again, and it will be great!

d1sam (approved):
Well done!

PR #903, opened 2024-03-21
implement Injector

StarAntonU:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Component {
```
kerrrusha, line 7: dont forget about target annotation

kerrrusha (approved):

PR #908, opened 2024-03-24
Implement dependency injection with @Component and @Inject annotations

VolandevlodD:
Implemented dependency injection functionality using @Component and @Inject annotations for instance creation and field initialization. Ensured that classes without the @Component annotation cannot be instantiated via the Injector, throwing an exception as required.


nklimovych (approved):
Nice Job!

PR #910, opened 2024-03-25
Task was completed

slizko1:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,69 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | InvocationTargetException
+                 | InstantiationException | IllegalAccessException e) {
```
nklimovych, line 65: It is better to replace many exceptions that have a common parent with a general parental exception.
```suggestion
        } catch (ReflectiveOperationException e) {
```

nklimovych (changes requested):
Looks good, left you one suggested change, as noted in the common mistakes file.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,67 @@
...
+    private Map<Class<?>, Object> instances = new HashMap<>();
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Class<?> clazz = interfaceClazz.isInterface() ? interfaceImplementations.get(interfaceClazz)
+                : interfaceClazz;
```
d1sam, line 28: let's extract this logic to the separate method `findImplementation`

d1sam (changes requested):
Good job, but there is one little change, that should be done!

PR #913, opened 2024-03-25
it throws an error with wrong class

MolchanovAlexander:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,86 @@
...
+            Object clazzImplementationInstance = null;
+            Field[] declaredFields = interfaceClazz.getDeclaredFields();
+            for (Field field : declaredFields) {
+                if (field.isAnnotationPresent(Inject.class)) {
+                    // create a new obj
+                    Object fieldInstance = getInstance(field.getType());
+                    //create an obj of interfaceClazz (or implementation class )
+                    clazzImplementationInstance = createNewInstance(clazz);
+                    //set `field type object` to `interfaceClazz obj`
+                    field.setAccessible(true);
```
fedaktaras, line 36: With no comments would be better.

src/main/java/mate.academy/service/FileReaderService.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 
+@Component
 public interface FileReaderService {
```
fedaktaras, line 7: I am not sure do we need @Component here.

src/main/java/mate.academy/service/FileReaderService.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 
+@Component
```
fedaktaras, line 6: I am not sure do we need it here.... I dont have it. works fine

fedaktaras (approved):
Can't be good reviewer because don't understand this topic good enough yet.

src/main/java/mate.academy/service/FileReaderService.java
karma-o: yes, interfaces do not need this annotation, because it is used to indicate that this **class** is an implementation that is suitable to be used in our injector, so we can remove it, from all interfaces

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Inject {
```
karma-o, line 7: add `@Target` for you annotations, as per the common mistakes

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,86 @@
...
+                 | IllegalAccessException e) {
+            throw new RuntimeException("Can't create new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> findImplementations = new HashMap<>();
+        findImplementations.put(ProductService.class, ProductServiceImpl.class);
+        findImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        findImplementations.put(ProductParser.class, ProductParserImpl.class);
```
karma-o, line 79: make it a class-level field, as it says in the common mistakes, and name it in a right way

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,86 @@
...
+        // create a new obj
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException
+                 | InvocationTargetException
+                 | InstantiationException
+                 | IllegalAccessException e) {
```
karma-o, line 70: check common mistakes

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,86 @@
...
+            throw new RuntimeException("Can't create new instance of " + clazz.getName());
```
karma-o, line 71: let's throw some custom exception here  instead of raw runtime

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,86 @@
...
+                            + interfaceClazz.getName());
+        }
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
+        // if we created obj let use it
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        // create a new obj
```
karma-o, line 61: if we name methods right we can avoid some comments in the code and improve readability
```suggestion
    private Object getOrCreateNewInstance(Class<?> clazz) {
        if (instances.containsKey(clazz)) {
            return instances.get(clazz);
        }
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,86 @@
...
+    private final Map<Class<?>, Object> instances = new HashMap<>();
+
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        if (interfaceClazz.isAnnotationPresent(Component.class)) {
+            Class<?> clazz = findImplementation(interfaceClazz);
```
karma-o, line 26: Flip this `if` to follow the return/fail early pattern and not overcomplicate `if/else`s
```suggestion
        if (!interfaceClazz.isAnnotationPresent(Component.class)) {
            throw new RuntimeException( // replace it with some custom one
                    "Injection failed, missing @Component annotation on the class "
                            + interfaceClazz.getName());
        }
        Class<?> clazz = findImplementation(interfaceClazz);
        // other main logic
```

karma-o (changes requested):
Please always check the common mistakes before pushing for review

PR #915, opened 2024-03-25
feat: first-try

rel1c-hub:
build&test success

### DIFF
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,13 +1,75 @@
 package mate.academy.lib;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.util.HashMap;
+import java.util.Map;
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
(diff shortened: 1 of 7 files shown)


HAIOVYI (approved):
GJ!

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,71 @@
...
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz,instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can`t create a new instance of " + clazz.getName());
```
karma-o, line 51: we should always add the original exception that we have caught to the runtime one, to keep the info and stacktrace. This is a point from common mistakes, please do not forget about it
Also, it's a good practice to use custom exceptions instead of raw Runtime when you can, same for the other exception that you throw above
```suggestion
            throw new RuntimeException("Can`t create a new instance of " + clazz.getName(), e);
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,71 @@
...
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = Map.of(
+                FileReaderService.class, FileReaderServiceImpl.class,
+                ProductParser.class, ProductParserImpl.class,
+                ProductService.class, ProductServiceImpl.class);
```
karma-o, line 59: Do we want to create a new map each time this method is called? Yevhenii already mentioned that it should be on the class level

karma-o (approved):
Approved but with several important comments, please take a look

PR #916, opened 2024-03-25
Initial commit just not to get deadline missed.

fedaktaras:
You can reject without reading. I just made test passed.

### DIFF
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -2,7 +2,6 @@ package mate.academy.lib;
 
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
-import java.lang.reflect.InvocationTargetException;
 import java.util.HashMap;
 import java.util.Map;
 import mate.academy.service.FileReaderService;
@@ -14,7 +13,7 @@ import mate.academy.service.impl.ProductServiceImpl;
 
 public class Injector {
     private static final Injector injector = new Injector();
-    private static Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
+    private static final Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
 
     static {
         interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
@@ -22,7 +21,7 @@ public class Injector {
         interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
     }
 
-    private Map<Class<?>, Object> instances = new HashMap<>();
+    private final Map<Class<?>, Object> instances = new HashMap<>();
 
     public static Injector getInjector() {
         return injector;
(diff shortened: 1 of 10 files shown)


Eidenn005 (changes requested):

yevheniisynytsia (changes requested):
Well done! Left some minor comments

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (interfaceClazz.isInterface()) {
+            Class<?> someClass = interfaceImplementations.get(interfaceClazz);
+            if (someClass == null) {
+                throw new RuntimeException("Not supported class");
```
karma-o, line 68: add which `interfaceClazz` to the ex message

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,80 @@
...
+        if (clazzImplementationInstance == null) {
+            clazzImplementationInstance = createNewInstance(clazz);
+        }
+        return clazzImplementationInstance;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
+        if (!interfaceImplementations.containsValue(clazz)) {
+            throw new RuntimeException("Unsupported");
+        }
```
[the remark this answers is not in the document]
karma-o: this is because you are not using you `Component` annotation anywhere, you should add validation in `findImplementation` if a class has it

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+        Field[] declaredFields = clazz.getDeclaredFields();
+        Object clazzImplementationInstance = createNewInstance(clazz);
+        for (Field field : declaredFields) {
+            if (field.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = getInstance(field.getType());
+                field.setAccessible(true);
+                try {
+                    field.set(clazzImplementationInstance, fieldInstance);
+                } catch (IllegalAccessException e) {
+                    throw new RuntimeException("Can't initialize field value");
```
karma-o, line 40: we should always add the exception that we have fought to save the info it contains
```suggestion
                    throw new RuntimeException("Can't initialize field value", e);
```
Also better to use some custom ex instead, and check other places for similar mistakes

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+
+    private final Map<Class<?>, Object> instances = new HashMap<>();
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+
```
karma-o, line 29:
```suggestion
```

karma-o (changes requested):

yevheniisynytsia (approved):
🔥

PR #919, opened 2024-03-25
jv-dependency-injection

bodya4243:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Component {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Component {
 }
```
kaiiiseeel, line 8: You need to add `@Target` as well (check common mistakes)

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Inject {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Inject {
 }
```
kaiiiseeel, line 8: Here you also need to add `@Target`

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,89 @@
...
+            return instances.get(clazz);
+        }
+
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (InstantiationException | InvocationTargetException | IllegalAccessException
+                 | NoSuchMethodException e) {
```
kaiiiseeel, line 57:
```suggestion
        } catch (ReflectiveOperationException e) {
```
Common mistakes again

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,89 @@
...
+                            interfaceImplementations.put(interfaceClazz, clazz);
+                        }
+                    }
+                } catch (ClassNotFoundException e) {
+                    throw new RuntimeException("there is no Implementations to this interface: "
+                            + interfaceClazz, e);
+                }
+            }
+        }
+        return interfaceImplementations.get(interfaceClazz);
```
kaiiiseeel, line 87: Looks interesting. I don't know if we need to do it this way (because there is a much simpler implementation), but ok

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,89 @@
...
+            instances.put(clazz, instance);
+            return instance;
+        } catch (InstantiationException | InvocationTargetException | IllegalAccessException
+                 | NoSuchMethodException e) {
+            throw new RuntimeException("can't create new instance of: " + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
```
kaiiiseeel, line 63: I think, we should create it as class-level variable, so we won't create new HashMap every time we calling this method

kaiiiseeel (changes requested):
LGTM! But you need to check common mistakes and make some improvements

src/main/java/mate.academy/lib/Component.java
karma-o: agree

src/main/java/mate.academy/lib/Inject.java
karma-o: agree

src/main/java/mate.academy/lib/Injector.java
karma-o: Looks very cool, good job!
An improvement idea that I have in mind for this is we can try to update this method and call it in the static initialization block, and try to scan and find implementations for the whole project there, to have that `interfaceImplementations` map filled before we even call this `findImplementation()`.
Also, there is no check that impl class is marked with component annotation, we should definitely add it somewhere here.
But of course, all of that is extra work and not necessary at all in this task, just for your free time :)

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,89 @@
...
+        String packageName = "mate.academy.service.impl";
+        String packagePath = packageName.replace('.', '/');
+
+        File packageDirectory = new File(Thread.currentThread()
+                .getContextClassLoader().getResource(packagePath).getFile());
+
+        for (File file : packageDirectory.listFiles()) {
+            if (file.isFile() && file.getName().endsWith(".class")) {
+                String className = packageName + "."
+                        + file.getName().replace(".class", "");
```
karma-o, line 73: lots of string literals, make them constants

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,89 @@
...
+                }
+            }
+        }
+        if (clazzImplementationInstance == null) {
+            clazzImplementationInstance = createNewInstance(clazz);
+        }
+        return clazzImplementationInstance;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
```
karma-o, line 46: Adds a bit more clarity
```suggestion
    private Object getOrCreateNewInstance(Class<?> clazz) {
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,89 @@
...
+            if (declaredField.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = getInstance(declaredField.getType());
+
+                clazzImplementationInstance = createNewInstance(clazz);
+
+                declaredField.setAccessible(true);
+                try {
+                    declaredField.set(clazzImplementationInstance, fieldInstance);
+                } catch (IllegalAccessException e) {
+                    throw new RuntimeException("Can't initialize field value. "
```
karma-o, line 34: I'd rather throw some custom exception, overall a good practice everywhere

src/main/java/mate.academy/service/impl/ProductParserImpl.java
```diff
@@ -1,20 +1,23 @@
...
     public static final int CATEGORY_POSITION = 2;
     public static final int DESCRIPTION_POSITION = 3;
     public static final int PRICE_POSITION = 4;
 
+    private final Product product = new Product();
+
     @Override
     public Product parse(String productInfo) {
         String[] data = productInfo.split(",");
-        Product product = new Product();
```
karma-o, line 17: why did you move it to the field? this is not a service class

karma-o (changes requested):
Impressive `findImplementation()` method, but do not forget about the common mistakes 😉

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,89 @@
...
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("can't create new instance of: " + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
```
karma-o, line 64: still missing check if `@Component` is present on the impl class that we have found, this is part of the requirements, so lets add it

karma-o (approved):
Approved, but you missed an important comment, so please take a look at it!

PR #922, opened 2024-03-26
My solution di

petro2061:
Injector class is implemented, which is used to initialize object fields. Corresponding annotations on classes and fields that need to be initialized have also been added. In the code, a check for marking the class with the corresponding @Component annotation has been added.

### DIFF
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,5 +1,17 @@
 package mate.academy.lib;
 
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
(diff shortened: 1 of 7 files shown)


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,78 @@
...
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        Field[] declaredFields = clazz.getDeclaredFields();
+        for (Field filed : declaredFields) {
+            if (filed.isAnnotationPresent(Inject.class)) {
+
```
Vasul-Pankiv, line 28:
```suggestion
```

Vasul-Pankiv (approved):
Just remove an empty line

PR #926, opened 2024-03-28
feat: add injector

oksana-miazina:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,80 @@
...
+    private Map<Class<?>, Object> instances = new HashMap<>();
+    private Map<Class<?>, Class<?>> interfaceImplementations;
+
+    {
+        interfaceImplementations = Map.of(
+                ProductService.class, ProductServiceImpl.class,
+                FileReaderService.class, FileReaderServiceImpl.class,
+                ProductParser.class, ProductParserImpl.class
+        );
+    }
```
karma-o, line 26:
```suggestion
    private final Map<Class<?>, Class<?>> interfaceImplementations = Map.of(
                ProductService.class, ProductServiceImpl.class,
                FileReaderService.class, FileReaderServiceImpl.class,
                ProductParser.class, ProductParserImpl.class
        );
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,80 @@
...
+
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException(
```
karma-o, line 35: it's a good practice to throw custom exceptions instead of raw runtime, same in other places

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,80 @@
...
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (interfaceClazz.isInterface()) {
+            return interfaceImplementations.get(interfaceClazz);
+        }
+        return interfaceClazz;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
```
karma-o, line 67: makes it a bit more clear on what does it do
```suggestion
    private Object getOrCreateNewInstance(Class<?> clazz) {
```

karma-o (approved):
Good job, only some minor comments!

PR #928, opened 2024-04-01
jv-dependency-injection-solution 1

DojiDoe:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
 public class Injector {
     private static final Injector injector = new Injector();
+    private static Map<Class<?>, Class<?>> interfaceImplementations;
+    private final Map<Class<?>, Object> instances = new HashMap<>();
 
     public static Injector getInjector() {
+        interfaceImplementations = Map.of(
+                FileReaderService.class, FileReaderServiceImpl.class,
+                ProductParser.class, ProductParserImpl.class,
+                ProductService.class, ProductServiceImpl.class);
```
kerrrusha, line 23: better replace this initialization
```suggestion
    private static Map<Class<?>, Class<?>> interfaceImplementations = Map.of(
                FileReaderService.class, FileReaderServiceImpl.class,
                ProductParser.class, ProductParserImpl.class,
                ProductService.class, ProductServiceImpl.class);
    private final Map<Class<?>, Object> instances = new HashMap<>();

    public static Injector getInjector() {
```

kerrrusha (approved):
overall is good

PR #929, opened 2024-04-04
jv-dependency-injection-resolved

margosalii:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+                 | IllegalAccessException | InvocationTargetException e) {
+            throw new RuntimeException("Can't create new instance of " + clazz.getName() + e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> intefraceImplementations = new HashMap<>();
+        intefraceImplementations.put(ProductService.class, ProductServiceImpl.class);
+        intefraceImplementations.put(ProductParser.class, ProductParserImpl.class);
+        intefraceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
```
olekskov, line 70: Please rework to initialize this map once. No need to create new map each time method is called

olekskov (changes requested):

okuzan (approved):

PR #933, opened 2024-04-10
do HW

RetKinf:
don't understand why the main isn't working

### DIFF
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,13 +1,76 @@
 package mate.academy.lib;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.util.HashMap;
+import java.util.Map;
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private Map<Class<?>, Object> instances = new HashMap<>();
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        Field[] declaredFields = interfaceClazz.getDeclaredFields();
+        for (Field field : declaredFields) {
+            if (!field.getClass().isAnnotationPresent(Component.class)) {
+                throw new RuntimeException("Injection failed, missing"
+                        + " @Component annotaion on the"
+                        + "class " + field.getClass());
+            }
+            if (!field.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = getInstance(field.getType());
+                clazzImplementationInstance = createNewInstance(clazz);
+                field.setAccessible(true);
+                try {
+                    field.set(clazzImplementationInstance, fieldInstance);
+                } catch (IllegalAccessException e) {
+                    throw new RuntimeException("Can't initialized filed value. Clazz: "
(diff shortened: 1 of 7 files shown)


src/main/java/mate.academy/service/impl/FileReaderServiceImpl.java
```diff
@@ -4,12 +4,15 @@
...
 import java.util.List;
+import mate.academy.lib.Component;
 import mate.academy.service.FileReaderService;
 
+@Component
 public class FileReaderServiceImpl implements FileReaderService {
     @Override
     public List<String> readFile(String fileName) {
         try {
+            System.out.println(new File(fileName).toPath());
```
okuzan, line 15: don't forget to remove such things

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>(
+                Map.of(
+                        FileReaderService.class, FileReaderServiceImpl.class,
```
okuzan, line 67: make it a class field to avoid needless recreation

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+        Object clazzImplementationInstance = null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        Field[] declaredFields = interfaceClazz.getDeclaredFields();
+        for (Field field : declaredFields) {
+            if (!field.getClass().isAnnotationPresent(Component.class)) {
+                throw new RuntimeException("Injection failed, missing"
+                        + " @Component annotaion on the"
+                        + "class " + field.getClass());
+            }
+            if (!field.isAnnotationPresent(Inject.class)) {
```
okuzan, line 32:
```suggestion
        if (!clazz.isAnnotationPresent(Component.class)) {
            throw new RuntimeException("Class " + clazz.getName()
                    + " doesn't have Component annotation");
        }
        Field[] declaredFields = clazz.getDeclaredFields();
        for (Field field : declaredFields) {
            if (field.isAnnotationPresent(Inject.class)) {
```

okuzan (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+        interfaceImplementations.putAll(
+                Map.of(
+                        FileReaderService.class, FileReaderServiceImpl.class,
+                        ProductParser.class,ProductParserImpl.class,
+                        ProductService.class, ProductServiceImpl.class
+        ));
+        if (interfaceClazz.isInterface()) {
+            return interfaceImplementations.get(interfaceClazz);
+        }
+        return interfaceClazz;
```
olekskov, line 74: There is no need to create new map for each call of findImplementation() method. Your map contains static information, create it only once on class level and reuse

olekskov (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private Map<Class<?>, Object> instances = new HashMap<>();
+    private Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>(
+            Map.of(
+                    FileReaderService.class, FileReaderServiceImpl.class,
+                    ProductParser.class,ProductParserImpl.class,
+                    ProductService.class, ProductServiceImpl.class
+    ));
```
olekskov, line 22:
```suggestion
    private Map<Class<?>, Class<?>> interfaceImplementations = Map.of(
                    FileReaderService.class, FileReaderServiceImpl.class,
                    ProductParser.class,ProductParserImpl.class,
                    ProductService.class, ProductServiceImpl.class
    );
```

olekskov (approved):
Good job

PR #937, opened 2024-04-21
Dependency Injection

MaksimSokur:
1. Create new annotation Component
2. Check in class Injector - annotation Component is present.
3. All test checked.

### DIFF
--- a/src/main/java/mate.academy/lib/Component.java
+++ b/src/main/java/mate.academy/lib/Component.java
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Component {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Component {
 }
--- a/src/main/java/mate.academy/lib/Inject.java
+++ b/src/main/java/mate.academy/lib/Inject.java
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Inject {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Inject {
 }
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,13 +1,83 @@
 package mate.academy.lib;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.util.HashMap;
+import java.util.Map;
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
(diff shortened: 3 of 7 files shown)


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Component {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
```
okuzan, line 6: add Target - checklist issue

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Inject {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
```
okuzan, line 6: add Target - checklist issue

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,83 @@
...
+        } catch (InvocationTargetException | InstantiationException | IllegalAccessException e) {
+            throw new RuntimeException("Can`t create new instance of " + clazz.getName(), e);
+        }
+    }
+
+    public Class<?> findImplementationInstance(Class<?> interfaceClass) {
+        Map<Class<?>, Class<?>> searchInterfaceImplMap = Map.of(
+                FileReaderService.class, FileReaderServiceImpl.class,
+                ProductService.class, ProductServiceImpl.class,
+                ProductParser.class, ProductParserImpl.class);
```
okuzan, line 80: extract to a class scope, to avoid needless recreation per each method call

okuzan (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,86 @@
...
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private final Map<Class<?>, Object> instances = new HashMap<>();
+    private final Map<Class<?>, Class<?>> searchInterfaceImplMap = new HashMap<>();
+
+    private Injector() {
+        searchInterfaceImplMap.put(FileReaderService.class, FileReaderServiceImpl.class);
+        searchInterfaceImplMap.put(ProductParser.class, ProductParserImpl.class);
+        searchInterfaceImplMap.put(ProductService.class, ProductServiceImpl.class);
```
okuzan, line 23: could use map.of

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,86 @@
...
+        Constructor<?> constructor;
+
+        try {
+            constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException e) {
+            throw new RuntimeException("No such method." + e);
+        } catch (InvocationTargetException | InstantiationException | IllegalAccessException e) {
```
okuzan, line 78: checklist issue, please always read it before submitting your solution 😥
image

okuzan (changes requested):

PR #938, opened 2024-04-22
Implemented jv-dependency-injection

ynik404:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,70 @@
...
+        if (clazzImplementationInstance == null) {
+            clazzImplementationInstance = createNewInstance(clazz);
+        }
+        return clazzImplementationInstance;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Can't create an instance " + clazz.getName() + " class");
+        }
```
MAMentorFX, line 46: too late to validate here. You can do it sooner, even before the forloop

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,70 @@
...
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create new instance of " + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = Map.of(
+                FileReaderService.class, FileReaderServiceImpl.class,
+                ProductParser.class, ProductParserImpl.class,
+                ProductService.class, ProductServiceImpl.class);
```
MAMentorFX, line 64: mkae it constant, do not create this map each time you call findIm0plementation()

MAMentorFX (changes requested):

okuzan (approved):

PR #946, opened 2024-04-29
solved task

Leanchers:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+            Object instance = constructor.newInstance();
+            instances.put(clazz,instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementation = new HashMap<>();
```
Elena-Bruyako, line 64: It's better to create class-level variable and use static block for initialisation 
```
static {
        interfaceImplementations = Map.of(
```

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #955, opened 2024-05-13
done

ChabVlad:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Cannot create new instance of " + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
```
kerrrusha, line 69: this map should be class-level field in order to be initialized only once

kerrrusha (changes requested):

src/main/java/mate.academy/lib/Injector.java
MentorMA: not fixed

MentorMA (changes requested):

MentorMA (approved):

PR #963, opened 2024-05-28
Created Injector Implementation with @Component(Class annotation) and…

VadymOziPantielieienko:
… @Inject(Field annotation).

### DIFF
--- a/src/main/java/mate.academy/lib/Component.java
+++ b/src/main/java/mate.academy/lib/Component.java
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Component {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Component {
 }
--- a/src/main/java/mate.academy/lib/Inject.java
+++ b/src/main/java/mate.academy/lib/Inject.java
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Inject {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Inject {
 }
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,13 +1,75 @@
 package mate.academy.lib;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.util.HashMap;
+import java.util.Map;
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private final Map<Class<?>, Object> instances = new HashMap<>();
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplInstance = null;
+        Class<?> clazz = findImpl(interfaceClazz);
+        Field[] declaredFields = clazz.getDeclaredFields();
+        for (Field field : declaredFields) {
+            if (field.isAnnotationPresent(Inject.class)) {
+                Object fieldInstance = getInstance(field.getType());
+                clazzImplInstance = createNewInstance(clazz);
+                field.setAccessible(true);
+                try {
+                    field.set(clazzImplInstance, fieldInstance);
+                } catch (IllegalAccessException e) {
+                    throw new RuntimeException("Can`t initialize field value. "
+                            + "Class: " + clazz.getName() + " Field: " + field.getName());
+                }
+            }
+        }
+        if (clazzImplInstance == null) {
+            clazzImplInstance = createNewInstance(clazz);
+        }
+        return clazzImplInstance;
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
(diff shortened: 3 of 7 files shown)


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Component {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
```

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Inject {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+    private final Map<Class<?>, Object> instances = new HashMap<>();
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Object clazzImplInstance = null;
+        Class<?> clazz = findImpl(interfaceClazz);
```
Elena-Bruyako, line 25: Do we need to check ?
```suggestion
        Class<?> clazz = findImpl(interfaceClazz);
        if (!clazz.isAnnotationPresent(Component.class)) { 
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,75 @@
...
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Can`t create new instance of this class "
+                    + clazz.getName());
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException | IllegalAccessException
```
Elena-Bruyako, line 59:
```suggestion
        } catch (ReflectiveOperationException e) {
```

Elena-Bruyako (changes requested):

src/main/java/mate.academy/lib/Injector.java
VadymOziPantielieienko: yeah, the task asks us to do this, or u mean that I should change it and do it differently?

"If the `Component` annotation is missing above the class, we shouldn't be able to create an instance of this class in Injector, and we should throw an exception"

VadymOziPantielieienko: feels like this is the simplest way to stop the process of creating new instances for classes that are not permitted by annotation

Elena-Bruyako: It's better to check `isAnnotationPresent(Component.class)` after `Class<?> clazz = findImpl(interfaceClazz);`
to stop the process and don't implement next logic

Elena-Bruyako (changes requested):
Answered to your comment

VadymOziPantielieienko:
[Fixed isAnnotationPresent for Component class

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can`t create new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImpl(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
```
MAMentorFX, line 67: make it class constant, you dont want to recreate this everytime you call `findImpl`

MAMentorFX (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can`t create new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImpl(Class<?> interfaceClazz) {
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
```
kerrrusha, line 67: bootstraping `interfaceImplementations` map should be done only once, not every time this method is being called

src/main/java/mate.academy/lib/Injector.java
kerrrusha: resolved

kerrrusha (changes requested):
almost perfect

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private static final Map<Class<?>, Object> instances = new HashMap<>();
+    private static final Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
```
MentorMateAcademy, line 17: Use map.of  for implementation of your realizations

src/main/java/mate.academy/lib/Injector.java
MentorMateAcademy: not here

MentorMateAcademy (changes requested):

PR #964, opened 2024-05-30
Initial commit: hw-dependency-injection solution

OleksiiLytvynenko:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,80 @@
...
+            throw new RuntimeException("Can't create a new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = Map.of(
+                FileReaderService.class, FileReaderServiceImpl.class,
+                ProductParser.class, ProductParserImpl.class,
+                ProductService.class, ProductServiceImpl.class
+        );
```
kerrrusha, line 70: should be class-level and created once, not each time this method is being called

kerrrusha (changes requested):

kerrrusha (approved):

PR #971, opened 2024-06-26
Develop

mgsorigo:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,87 @@
...
+            LOGGER.log(Level.SEVERE, "Injection failed for " + interfaceClass.getName()
+                    + " due to " + e.getClass().getSimpleName(), e);
+            throw new RuntimeException(
+                    "Injection failed for " + interfaceClass.getName()
+                            + " due to " + e.getClass().getSimpleName(), e
+            );
+        }
+    }
+
+    private Object createInstance(Class<?> clazz) throws ReflectiveOperationException {
```
MateuszRostek, line 60: You should handle the exception inside this method, do not throw it in the method signature.

src/main/java/mate.academy/lib/Injector.java
MateuszRostek, line 1: Please take a look at this solution and compare it with yours, for example the `createInstance` method.

```java
public class Injector {
    private final Map<Class<?>, Class<?>> implementations;
    private final Map<Class<?>, Object> instances;

    public Injector() {
        implementations = Map.of(
        );
        instances = new HashMap<>();
    }

        return injector;
    }

        Object clazzImplementationInstance = null;
        Class<?> clazz = getImplementation(interfaceClazz);

            throw new RuntimeException("Missing @Component annotation in class: "
                    + clazz.getName() + ". Injection failed!");
        }
        for (Field field : clazz.getDeclaredFields()) {
            if (field.isAnnotationPresent(Inject.class)) {
                Object fieldInstance = getInstance(field.getType());
                clazzImplementationInstance = createNewInstance(clazz);
                field.setAccessible(true);
                try {
                    field.set(clazzImplementationInstance, fieldInstance);
                } catch (IllegalAccessException e) {
                    throw new RuntimeException("Can't initialize field value. Class "
                            + clazz.getName() + ". Field " + field.getName(), e);
                }
            }
        }
        if (clazzImplementationInstance == null) {
            clazzImplementationInstance = createNewInstance(clazz);
        }
        return clazzImplementationInstance;
    }

    private Object createNewInstance(Class<?> clazz) {
            return instances.get(clazz);
        }
        Constructor<?> constructor;
        try {
            constructor = clazz.getConstructor();
            Object newInstance = constructor.newInstance();
            instances.put(clazz, newInstance);
            return newInstance;
            throw new RuntimeException("Can't create a new instance of: " + clazz.getName(), e);
        }
    }

    private Class<?> getImplementation(Class<?> interfaceClazz) {
            return implementations.get(interfaceClazz);
        }
        return interfaceClazz;
    }
}
```

MateuszRostek (changes requested):
Nice try, but there are some changes required.

src/main/java/mate.academy/Main.java
MateuszRostek, line 1: What is the point of catching this RuntimeException? Just do it as I told you in the previous review.

MateuszRostek (approved):
Good job! See one minor advice.

PR #978, opened 2024-07-08
hw done

BelitskyiAlexandr:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,10 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Component {
 
 }
+
```
sarakhmen, line 10:
```suggestion
```

src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,10 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
 public @interface Component {
```
sarakhmen, line 7: add Target annotation as well, fix the same issue in all places

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+        }
+
+        for (Class<?> clazz : getAllClasses()) {
+            if (interfaceClass.isAssignableFrom(clazz)
+                    && clazz.isAnnotationPresent(Component.class)) {
+                try {
+                    T instance = (T) createInstance(clazz);
+                    instances.put(interfaceClass, instance);
+                    return instance;
+                } catch (Exception e) {
```
sarakhmen, line 32: please use more narrow exception class instead of Exception

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Class " + clazz.getName()
+                    + " is not annotated with @Component");
+        }
+
+        T instance;
+        try {
+            instance = clazz.getDeclaredConstructor().newInstance();
+        } catch (InstantiationException | IllegalAccessException
+                 | InvocationTargetException | NoSuchMethodException e) {
```
sarakhmen, line 50: please use their common ancestor

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,77 @@
...
+
+        return instance;
+    }
+
+    private List<Class<?>> getAllClasses() {
+        return List.of(
+                FileReaderServiceImpl.class,
+                ProductParserImpl.class,
+                ProductServiceImpl.class
+        );
```
sarakhmen, line 75: move to the class constants

sarakhmen (changes requested):
let's improve your solution :)

src/main/java/mate.academy/lib/Injector.java
BelitskyiAlexandr: image

But their common ancestor is 'Throwable': NoSuchMethodException is a subclass of 'Error', but InvocationTargetException is a subclass of 'Exception'

src/main/java/mate.academy/lib/Injector.java
BelitskyiAlexandr: But why? It is the only place, where it can be used in this class

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,81 @@
...
 public class Injector {
-    private static final Injector injector = new Injector();
+    private static final Injector INJECTOR = new Injector();
+    private static final FileReaderServiceImpl FILE_READER_SERVICE =
+            new FileReaderServiceImpl();
+    private static final ProductParserImpl PRODUCT_PARSER =
+            new ProductParserImpl();
+    private static final ProductServiceImpl PRODUCT_SERVICE =
+            new ProductServiceImpl();
+    private final Map<Class<?>, Object> instances = new HashMap<>();
```
MentorMateAcademy, line 19: you can init implementations here

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,81 @@
...
 
-    public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+    private List<Class<?>> getAllClasses() {
+        return List.of(
+                FILE_READER_SERVICE.getClass(),
+                PRODUCT_PARSER.getClass(),
+                PRODUCT_SERVICE.getClass()
+        );
     }
```
MentorMateAcademy, line 80: don`t need

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,81 @@
...
+
+        for (Field field : clazz.getDeclaredFields()) {
+            if (field.isAnnotationPresent(Inject.class)) {
+                try {
+                    Class<?> fieldType = field.getType();
+                    Object fieldInstance = getInstance(fieldType);
+                    field.setAccessible(true);
+                    field.set(instance, fieldInstance);
+                } catch (IllegalAccessException e) {
+                    throw new RuntimeException(e);
```
MentorMateAcademy, line 66: add message

MentorMateAcademy (changes requested):

src/main/java/mate.academy/lib/Injector.java
MAMentorFX: Because List.of() creates new ArrayList everytime it's called. We don't want that. So we create this list only once

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,72 @@
...
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
-    private static final Injector injector = new Injector();
+    private static final Injector INJECTOR = new Injector();
+    private final Map<Class<?>, Object> instances = new HashMap<>(
+            Map.of(
+                    FileReaderServiceImpl.class, new FileReaderServiceImpl(),
+                    ProductParserImpl.class, new ProductParserImpl(),
+                    ProductServiceImpl.class, new ProductServiceImpl()
```
MAMentorFX, line 16:
```suggestion
                    FileReaderService.class, new FileReaderServiceImpl(),
                    ProductParser.class, new ProductParserImpl(),
                    ProductService.class, new ProductServiceImpl()
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,72 @@
...
+    );
 
     public static Injector getInjector() {
-        return injector;
+        return INJECTOR;
+    }
+
+    public <T> T getInstance(Class<T> interfaceClass) {
+        if (instances.containsKey(interfaceClass)) {
+            return interfaceClass.cast(instances.get(interfaceClass));
```
MAMentorFX, line 26:
```suggestion
            return instances.get(interfaceClass);
```
by interface you get an implementation, no need to cast anything

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,72 @@
...
+        throw new RuntimeException("No implementation found for " + interfaceClass.getName());
     }
 
-    public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+    private <T> T createInstance(Class<T> clazz) {
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("Class " + clazz.getName()
+                    + " is not annotated with @Component");
+        }
```
MAMentorFX, line 48: I dont think you'll ever enter this if, because of 31 line. I suggest you check if it's component before 29 line

MAMentorFX (changes requested):

src/main/java/mate.academy/lib/Injector.java
BelitskyiAlexandr: image

MentorMateAcademy (approved):

PR #983, opened 2024-07-29
implemented own Injector.java;

aksoli666:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
Elena-Bruyako, line 6: Add Target annotation

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
Elena-Bruyako, line 6: The same

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,90 @@
...
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException
+                 | InvocationTargetException
+                 | InstantiationException
+                 | IllegalAccessException e) {
```
Elena-Bruyako, line 86:
```suggestion
        } catch (ReflectiveOperationException e) {
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,90 @@
...
+                            + " of class " + interfaceClazz.getName(), e);
+                }
+            }
+        }
+        instances.put(interfaceClazz, classImplementationInstance);
+        return classImplementationInstance;
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> implementations = new HashMap<>();
```
Elena-Bruyako, line 50: It's better to create `private static final Map<Class<?>, Class<?>> interfaceImplementations;`
and initialise it in static block `interfaceImplementations = Map.of(...`

Elena-Bruyako (changes requested):

okuzan (approved):

PR #987, opened 2024-08-04
Implemented an injector

sovisou:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can`t create a new instance of: " + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImpl(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImpl = Map.of(
```
Elena-Bruyako, line 64: Make it class-level and use static block for initialisation

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #988, opened 2024-08-05
Implement logic injector

Roma1990Melnichuk:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Component {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
```
okuzan, line 6: common mistakes checklist - add Target

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,84 @@
...
+                }
+            }
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClass) {
+        Map<Class<?>, Class<?>> interfaceToImplementation = new HashMap<>();
+        interfaceToImplementation.put(ProductParser.class, ProductParserImpl.class);
+        interfaceToImplementation.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceToImplementation.put(ProductService.class, ProductServiceImpl.class);
```
okuzan, line 77: use Map.of, should be a field

okuzan (changes requested):

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Inject {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
```
Elena-Bruyako, line 6: Add `@Target`

sarakhmen (approved):

PR #989, opened 2024-08-08
Implement Injector with @Component and @Inject annotations for dependency injection

smagles:
…ency injection

### DIFF
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,5 +1,18 @@
 package mate.academy.lib;
 
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.util.HashMap;
+import java.util.Map;
+
 public class Injector {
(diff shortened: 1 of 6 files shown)


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private final Map<Class<?>, Object> instances = new HashMap<>();
+    private final Map<Class<?>, Class<?>> implementationMap = Map.of(
+            ProductParser.class, ProductParserImpl.class,
+            ProductService.class, ProductServiceImpl.class,
+            FileReaderService.class, FileReaderServiceImpl.class
+    );
```
MateuszRostek, line 21: Let's move it to a constructor, for example like this:

```java
    private final Map<Class<?>, Class<?>> implementations;
    private final Map<Class<?>, Object> instances;

    public Injector() {
        implementations = Map.of(
        );
        instances = new HashMap<>();
    }
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
+        if (instances.containsKey(clazz)) {
+            return instances.get(clazz);
+        }
+        Constructor<?> constructor;
+        try {
+            constructor = clazz.getConstructor();
+            Object object = constructor.newInstance();
```
MateuszRostek, line 61: Use more informative names, for example:
```suggestion
            Object newInstance = constructor.newInstance();
```

MateuszRostek (approved):
Good job! See minor advice and implement it, approved!

PR #994, opened 2024-08-25
 implemented an injector

svitlanagalicka:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,73 @@
...
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can not create a new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = Map.of(
```
Elena-Bruyako, line 64: it's better to make it class-level variable and initialiaze in static block

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #995, opened 2024-08-29
Dependency Injection Impl

dmarmul:
create implementation of Injector class

### DIFF
--- a/src/main/java/mate.academy/lib/Component.java
+++ b/src/main/java/mate.academy/lib/Component.java
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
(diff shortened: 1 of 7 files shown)


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
Elena-Bruyako, line 6: add Target

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #996, opened 2024-08-30
Implementation dependency injection

4Vitalii5:


src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
Elena-Bruyako, line 6: Add Target annotaiton

Elena-Bruyako (changes requested):

src/main/java/mate.academy/lib/Inject.java
4Vitalii5: Thank you.

okuzan (approved):

PR #999, opened 2024-08-31
commit Component

nplastun:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,66 @@
...
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private final Map<Class<?>, Class<?>> interfaceImplementations;
+    private final Map<Class<?>, Object> instances = new HashMap<>();
+
+    private Injector() {
```
Elena-Bruyako, line 18:
```suggestion
    static {
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,66 @@
...
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Class<?> implClass = interfaceImplementations.get(interfaceClazz);
+
+        if (implClass == null) {
+            throw new RuntimeException(
```
Elena-Bruyako, line 34:
```suggestion
            implClass = createNewInstance(clazz);
```

create method

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,66 @@
...
+        );
+    }
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Class<?> implClass = interfaceImplementations.get(interfaceClazz);
```
Elena-Bruyako, line 31: divide logic
create `private static Class<?> getImplementation(Class<?> interfaceClazz) {
`

Elena-Bruyako (changes requested):

src/main/java/mate.academy/lib/Injector.java
okuzan: not fixed
you can just make it a static field and init at declaration

src/main/java/mate.academy/lib/Injector.java
okuzan: not fixed

okuzan (changes requested):
some comments not fixed

okuzan (approved):
well done

PR #1001, opened 2024-09-03
JavaDependencyInjection

VladSlob:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,8 @@
 package mate.academy.lib;
 
-public @interface Component {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Component {
```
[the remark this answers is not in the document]
okuzan: According to the checklist, custom annotations should have a @Target annotation to specify where this annotation can be used. For example, if @Component is meant to be used on classes, you should add @Target(ElementType.TYPE).

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,65 @@
...
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (interfaceClazz.equals(ProductService.class)) {
+            return ProductServiceImpl.class;
+        } else if (interfaceClazz.equals(ProductParser.class)) {
+            return ProductParserImpl.class;
+        } else if (interfaceClazz.equals(FileReaderService.class)) {
+            return FileReaderServiceImpl.class;
+        }
```
okuzan, line 38: create a hashmap field

okuzan (changes requested):

src/main/java/mate.academy/service/ProductParser.java
```diff
@@ -3,12 +3,5 @@
...
 public interface ProductParser {
-    /**
-     * Parse the productInfo and return created {@link Product} object.
-     * The `productInfo` could be received from any source.
-     * For example from file with all products stored in the shop.
-     * @param productInfo - the information about a product
-     * @return newly create {@link Product} object based on the `productInfo` input.
-     */
-    Product parse(String productInfo);
+    Product parse(String line);
```
Elena-Bruyako, line 6: revert changes

src/main/java/mate.academy/service/impl/FileReaderServiceImpl.java
```diff
@@ -4,15 +4,19 @@
...
+import mate.academy.lib.Component;
 import mate.academy.service.FileReaderService;
 
+@Component
 public class FileReaderServiceImpl implements FileReaderService {
     @Override
     public List<String> readFile(String fileName) {
         try {
-            return Files.readAllLines(new File(fileName).toPath());
+            return Files.lines(new File(fileName).toPath())
```
Elena-Bruyako, line 16: revert changes

src/main/java/mate.academy/service/impl/FileReaderServiceImpl.java
```diff
@@ -4,15 +4,19 @@
...
+                    .collect(Collectors.toList());
         } catch (IOException e) {
-            throw new RuntimeException("Can't read file: " + fileName, e);
+            throw new RuntimeException("Can't read file " + fileName, e);
```
Elena-Bruyako, line 19: revert changes

src/main/java/mate.academy/service/impl/ProductServiceImpl.java
```diff
@@ -2,18 +2,31 @@
...
 public class ProductServiceImpl implements ProductService {
+    @Inject
     private ProductParser productParser;
+    @Inject
     private FileReaderService fileReaderService;
 
     @Override
-    public List<Product> getAllFromFile(String filePath) {
-        return fileReaderService.readFile(filePath)
+    public List<Product> getAllProducts() {
```
Elena-Bruyako, line 20: revert changes

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,40 @@
...
+    }
+
     public static Injector getInjector() {
         return injector;
     }
 
-    public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+    public <T> T getInstance(Class<T> interfaceClazz) {
+        Class<?> implClass = interfaceToImplementation.get(interfaceClazz);
```
Elena-Bruyako, line 30: This is not full implementation os Injector

The convention is to use Component on the classes that are used for instance creation (usually these are interface implementations) and Inject to mark fields that need to be initialized. If the Component annotation is missing above the class, we shouldn't be able to create an instance of this class in Injector, and we should throw an exception

Don't forget about annotations checks in your Injector class.
Let's check instances map before new instance creation.

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1004, opened 2024-09-12
Solution

MarcusPy:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+                    field.setAccessible(true);
+                    field.set(clazzImplementationInstance, fieldInstance);
+                } catch (IllegalAccessException e) {
+                    throw new IllegalStateException("Can`t initialize field value. "
+                            + "Class: " + clazz.getName()
+                            + ". Field: " + field.getName(), e);
+                }
+            }
+        }
+        if (clazzImplementationInstance == null) {
```
slade13, line 48: You're creating an instance of 'clazz' before checking if it's null. This can lead to NullPointerException if 'clazz' is null.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+    private static final Map<Class<?>, Class<?>> interfaceImplementations = Map.of(
```
slade13, line 16: The current implementation of 'interfaceImplementations' map does not support multiple implementations of a single interface. Consider using a different data structure or strategy to support this.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Class<?> clazz = findImplementation(interfaceClazz);
+        Component annotation = clazz.getAnnotation(Component.class);
+        if (annotation == null) {
+            throw new IllegalArgumentException("Injection failed"
+                    + ", missing @Component annotation on the class " + clazz.getName());
+        }
+        Object clazzImplementationInstance = createNewInstance(clazz);;
```
slade13, line 33: There is a redundant semicolon at the end of the line.

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new IllegalStateException("Can`t create a new instance of " + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceOrClazz) {
```
slade13, line 68: 'findImplementation' method currently only supports interfaces. It will not return correct implementation if a class is passed as an argument.

slade13 (approved):
Good job! There are some comments left but they do not interfere with task description or common rules.

PR #1008, opened 2024-09-16
Injector implementation

dlvsn:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,80 @@
...
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create a new instance of" + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = Map.of(
+                ProductService.class, ProductServiceImpl.class,
+                FileReaderService.class, FileReaderServiceImpl.class,
+                ProductParser.class, ProductParserImpl.class
```
okuzan, line 73: make it a class field, don't create it on a method level

okuzan (changes requested):

Elena-Bruyako (approved):

PR #1009, opened 2024-09-20
Implemented dependency injection with @Component and @Inject annotati…

Pancher07:
…ons.

Added an Injector to handle object instantiation based on @Component annotations. Classes without @Component throw an exception when instantiation is attempted. @Inject marks fields for initialization. Configured @Retention for annotation visibility with reflection.


okuzan (approved):
Good job

PR #1040, opened 2024-11-14
implemented field dependency injection

Nazar1106:


src/main/java/mate.academy/Main.java
```diff
@@ -6,9 +6,7 @@
 import mate.academy.service.ProductService;
 
 public class Main {
-
     public static void main(String[] args) {
-        // Please test your Injector here. Feel free to push this class as a part of your solution
         Injector injector = Injector.getInjector();
         ProductService productService = null;
```
Elena-Bruyako, line 11: Please test your Injector here. Feel free to push this class as a part of your solution

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
Elena-Bruyako, line 6: Add Target

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+        }
+        try {
+            Constructor<?> constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+            instances.put(clazz, instance);
+            return instance;
+        } catch (NoSuchMethodException
+                 | InstantiationException
+                 | IllegalAccessException
+                 | InvocationTargetException e) {
```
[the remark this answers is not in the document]
Elena-Bruyako: fix comment

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+            throw new RuntimeException("Can't create new instance of " + clazz.getName());
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementation = new HashMap<>();
```
Elena-Bruyako, line 64: it's better to create class-level variable and use Map.of(). It could be in static block of initialization

Elena-Bruyako (changes requested):

PR #1041, opened 2024-11-15
Solution

vsupruniuk:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,95 @@
...
+            constructor = clazz.getConstructor();
+            Object instance = constructor.newInstance();
+
+            instances.put(clazz, instance);
+
+            return instance;
+        } catch (NoSuchMethodException
+            | InvocationTargetException
+            | InstantiationException
+            | IllegalAccessException e
```
liliia-ponomarenko, line 76: you can replace it using ReflectiveOperationException

src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,7 @@
 package mate.academy.lib;
 
-public @interface Component {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
-}
+@Retention(RetentionPolicy.RUNTIME)
```
liliia-ponomarenko, line 6: better add Target

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,7 @@
 package mate.academy.lib;
 
-public @interface Inject {
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
-}
+@Retention(RetentionPolicy.RUNTIME)
```
liliia-ponomarenko, line 6: the same

liliia-ponomarenko (approved):
Good job! A few comments ;)

PR #1043, opened 2024-11-16
add solution

IhorRome:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
+                    + implementationClazz.getName(), e);
+        }
+
+        return instanceMap.get(implementationClazz);
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (interfaceClazz.isInterface()) {
+            return classImplMap.get(interfaceClazz);
+        } else {
```
okuzan, line 72:
```suggestion
        }
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,76 @@
...
 public class Injector {
     private static final Injector injector = new Injector();
+    private static final Map<Class<?>, Object> instanceMap = new HashMap<>();
+    private static final Map<Class<?>, Class<?>> classImplMap = new HashMap<>();
+
+    static {
+        classImplMap.put(FileReaderService.class, FileReaderServiceImpl.class);
+        classImplMap.put(ProductParser.class, ProductParserImpl.class);
+        classImplMap.put(ProductService.class, ProductServiceImpl.class);
+    }
```
okuzan, line 23: You can just init with Map.of(

okuzan (approved):

PR #1052, opened 2024-12-03
final

Nikolya777777777777:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,87 @@
...
+        } catch (InvocationTargetException f) {
+            throw new RuntimeException("Can't find instance");
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementations = new HashMap<>();
+        interfaceImplementations.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementations.put(ProductParser.class, ProductParserImpl.class);
+        interfaceImplementations.put(ProductService.class, ProductServiceImpl.class);
```
sarakhmen, line 81: this map is created and configured each time the method is called. Can you move the map with its content to the constants?

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,87 @@
...
+            return instance;
+        } catch (NoSuchElementException e) {
+            throw new RuntimeException("There is no a such element");
+        } catch (NoSuchMethodException x) {
+            throw new RuntimeException("There is no a such method");
+        } catch (InstantiationException z) {
+            throw new RuntimeException("Can't create instance");
+        } catch (IllegalAccessException y) {
+            throw new RuntimeException("Can't access to the class");
+        } catch (InvocationTargetException f) {
```
sarakhmen, line 72: It is better to replace many exceptions that have a common parent with a general parental exception

sarakhmen (changes requested):

src/main/java/mate.academy/service/ProductService.java
```diff
@@ -1,8 +1,10 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
```
JJJazl, line 7:
```suggestion
```

src/main/java/mate.academy/service/ProductParser.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
+import mate.academy.lib.Component;
 import mate.academy.model.Product;
 
+@Component
```
JJJazl, line 6:
```suggestion
```

src/main/java/mate.academy/service/FileReaderService.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
 import java.util.List;
+import mate.academy.lib.Component;
 
+@Component
```
JJJazl, line 6:
```suggestion
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+            ProductService.class, ProductServiceImpl.class
+    );
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        if (interfaceClazz.isAnnotationPresent(Component.class)) {
```
JJJazl, line 28: Please, see comments AI mentor. We use @Component only above classes, so we don't need to check if the interface has this annotation
```suggestion
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,74 @@
...
+            if (clazzImplementationInstance == null) {
+                clazzImplementationInstance = createNewInstance(clazz);
+            }
+            return clazzImplementationInstance;
+        } else {
+            throw new RuntimeException("This class don't have Component annotation");
+        }
+    }
+
+    private Object createNewInstance(Class<?> clazz) {
```
JJJazl, line 53: Here we need to check if class annotated with @Component and if it hasn't throw an exception with an informative message

JJJazl, line 53: We not always create a new instance in this method, so let's describe how it actually works
```suggestion
    private Object createNewInstanceOrGetFromCacheIfPresent(Class<?> clazz) {
```

JJJazl (changes requested):

src/main/java/mate.academy/service/ProductService.java
Nikolya777777777777: I can't delete this, because 2 tests fails without this annotation here

src/main/java/mate.academy/service/FileReaderService.java
Nikolya777777777777: the same here

src/main/java/mate.academy/lib/Injector.java
Nikolya777777777777: But tests are written for interfaces and if I will delete this, tests will fail

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,80 @@
...
+            ProductService.class, ProductServiceImpl.class
+    );
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        if (interfaceClazz.isAnnotationPresent(Component.class)) {
```
Elena-Bruyako, line 28: findImplementation
check if Component annotation is present under the clazz. If no - throw an exception
get declared fields and etc

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,80 @@
...
+                clazzImplementationInstance = createNewInstanceOrGetFromCacheIfPresent(clazz);
+            }
+            return clazzImplementationInstance;
+        } else {
+            throw new RuntimeException("This class don't have Component annotation");
+        }
+    }
+
+    private Object createNewInstanceOrGetFromCacheIfPresent(Class<?> clazz) {
+        if (!clazz.isAnnotationPresent(Component.class)) {
```
Elena-Bruyako, line 55: remove

src/main/java/mate.academy/service/FileReaderService.java
Elena-Bruyako: remove `@Component`

src/main/java/mate.academy/service/ProductParser.java
Elena-Bruyako, line 6: remove `@Component`

src/main/java/mate.academy/service/ProductService.java
Elena-Bruyako, line 7: remove `@Component`

Elena-Bruyako (changes requested):

sarakhmen (approved):

PR #1053, opened 2024-12-08
First try

YanHrdMkr:


src/main/java/mate.academy/lib/Component.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
Elena-Bruyako, line 6: add `@Target`

src/main/java/mate.academy/lib/Inject.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.lib;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
Elena-Bruyako, line 6: same here

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,63 @@
...
+            return implInstance;
+        } catch (Exception e) {
+            throw new RuntimeException("Can't create instance of class: "
+                    + interfaceClazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        if (interfaceClazz.isInterface()) {
+            if (interfaceClazz.equals(mate.academy.service.ProductService.class)) {
```
Elena-Bruyako, line 52: let's create Map with implementations (class-level variable)
you can use Map.of()

Elena-Bruyako (changes requested):

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,12 +1,18 @@
...
     private static final Injector injector = new Injector();
-    private final Map<Class<?>, Object> instances = new HashMap<>();
+
+    private static final Map<Class<?>, Class<?>> IMPLEMENTATIONS = Map.of(
+            mate.academy.service.ProductService.class,
+            mate.academy.service.impl.ProductServiceImpl.class,
+            mate.academy.service.ProductParser.class,
+            mate.academy.service.impl.ProductParserImpl.class,
+            mate.academy.service.FileReaderService.class,
+            mate.academy.service.impl.FileReaderServiceImpl.class
```
sarakhmen, line 14: why did you add full package names instead of importing these classes?

sarakhmen (changes requested):

sarakhmen (approved):

PR #1081, opened 2025-02-04
Solution

MaybeTI:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,72 @@
...
+                 | IllegalAccessException e) {
+            throw new RuntimeException("Can`t create a new instance of " + clazz.getName(), e);
+        }
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> interfaceImplementation = new HashMap<>();
+        interfaceImplementation.put(ProductService.class, ProductServiceImpl.class);
+        interfaceImplementation.put(FileReaderService.class, FileReaderServiceImpl.class);
+        interfaceImplementation.put(ProductParser.class, ProductParserImpl.class);
```
d1sam, line 66: do you really need to create the same map on each call of `findImplementation`?

d1sam (changes requested):
1 minor change is needed

PR #1105, opened 2025-03-27
implement DI

deez-nerdy:
implement Injector that uses @component and @inject annotations to initialize instances

### DIFF
--- a/src/main/java/mate.academy/lib/Injector.java
+++ b/src/main/java/mate.academy/lib/Injector.java
@@ -1,13 +1,88 @@
 package mate.academy.lib;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
+
+    private final Map<Class<?>, Object> instances = new HashMap<>();
 
     public static Injector getInjector() {
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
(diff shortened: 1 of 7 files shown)


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,88 @@
...
+                    throw new RuntimeException("Can't initialize field "
+                            + field.getName());
+                }
+            }
+        }
+        return classImplementationInstance;
+    }
+
+    private Class<?> findImplementation(Class<?> interfaceClass) {
+        Map<Class<?>, Class<?>> interfaceImplementation = Map.of(
```
Elena-Bruyako, line 55: it's better to make `interfaceImplementation` as class-level variable and use static block to initialise Map

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,88 @@
...
+            return instance;
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException("Can't create new instance of "
+                    + clazz.getName());
+        }
+    }
+
+    private void componentAnnotationChecker(Class<?> clazz) {
+        if (!clazz.isAnnotationPresent(Component.class)) {
+            throw new RuntimeException("There is no '@Component' annotation in class "
```
Elena-Bruyako, line 84:
```suggestion
            throw new RuntimeException("There is no Component annotation in class "
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,88 @@
...
+            throw new RuntimeException("There is no implementation for class "
+                    + interfaceClazz.getName());
+        }
+
+        componentAnnotationChecker(clazz);
+
+        Object classImplementationInstance = createNewInstance(clazz);
+
+        List<Field> fields = List.of(clazz.getDeclaredFields());
+
```
Elena-Bruyako, line 37: remove redundant empty lines
```suggestion
        componentAnnotationChecker(clazz);
        Object classImplementationInstance = createNewInstance(clazz);
        List<Field> fields = List.of(clazz.getDeclaredFields());
```

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1135, opened 2025-06-27
Implement custom dependency injection mechanism:

Mwxyz2022:
- configured @component and @inject annotations with runtime retention policy
- implemented the core logic for the injector class to create and manage instances
- added a caching mechanism to the injector to store and reuse created instances
- used reflection to instantiate objects based on their constructors
- enabled automatic dependency injection for fields annotated with @inject
- marked service implementations (FileReaderServiceImpl, ProductParserImpl, ProductServiceImpl) with the @Component annotation
- injected dependencies into ProductServiceImpl using the @Inject annotation
- updated the main method to retrieve the ProductService instance via the injector


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,68 @@
...
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
 
+    private final Map<Class<?>, Object> instances = new HashMap<>();
+    private final Map<Class<?>, Class<?>> interfaceImplementations;
+
+    private Injector() {
```
Elena-Bruyako: it's better to use static block of initialization

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,68 @@
...
         return injector;
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        if (instances.containsKey(interfaceClazz)) {
+            return instances.get(interfaceClazz);
+        }
+
+        Class<?> implementationClass = interfaceImplementations.get(interfaceClazz);
```
Elena-Bruyako, line 54: create separate method `getImplementation`

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,13 +1,68 @@
...
+
+        try {
+            Constructor<?> constructor = implementationClass.getConstructor();
+            Object instance = constructor.newInstance();
+
+            Field[] declaredFields = implementationClass.getDeclaredFields();
+            for (Field field : declaredFields) {
+                if (field.isAnnotationPresent(Inject.class)) {
+                    Object fieldInstance = getInstance(field.getType());
+                    field.setAccessible(true);
```
Elena-Bruyako, line 73: create method to create new instance through the constructor

Elena-Bruyako (changes requested):

liliia-ponomarenko (approved):
Good job!

PR #1138, opened 2025-07-24
Homework 1

Woterix:


src/main/java/mate.academy/lib/Injector.java
```diff
@@ -8,6 +19,48 @@ public static Injector getInjector() {
...
+                } catch (IllegalAccessException e) {
+                    throw new RuntimeException("Failed to inject field: " + field.getName(), e);
+                }
+            }
+        }
+        return instance;
+    }
+
+    private Class findImplementation(Class<?> interfaceClazz) {
+        Map<Class<?>, Class<?>> classClassMap = new HashMap<>();
```
Elena-Bruyako, line 41: it should be class level variable

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -1,5 +1,16 @@
...
+import java.util.Map;
+import mate.academy.service.FileReaderService;
+import mate.academy.service.ProductParser;
+import mate.academy.service.ProductService;
+import mate.academy.service.impl.FileReaderServiceImpl;
+import mate.academy.service.impl.ProductParserImpl;
+import mate.academy.service.impl.ProductServiceImpl;
+
 public class Injector {
     private static final Injector injector = new Injector();
```
Elena-Bruyako, line 15:
```suggestion
    private static final Injector injector = new Injector();
    private static final Map<Class<?>, Class<?>> interfaceImplementations;
    private Map<Class<?>, Object> instances = new HashMap<>();

    static {
        interfaceImplementations = Map.of(
                FileReaderService.class, FileReaderServiceImpl.class,
                ProductService.class, ProductServiceImpl.class,
                ProductParser.class, ProductParserImpl.class);
    }
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -8,6 +19,48 @@ public static Injector getInjector() {
...
+        for (Map.Entry<Class<?>, Class<?>> entry : classClassMap.entrySet()) {
+            if (entry.getKey().equals(interfaceClazz)) {
+                return entry.getValue();
+            }
+        }
+        return interfaceClazz;
+    }
+
+    private Object createObject(Class<?> clazz) {
+        if (!clazz.isAnnotationPresent(Component.class)) {
```
Elena-Bruyako, line 54:
```suggestion
        if (instances.containsKey(clazz)) {
            return instances.get(clazz);
        }
        try {
            Constructor<?> constructor = clazz.getConstructor();
            Object instance = constructor.newInstance();
            instances.put(clazz, instance);
            return instance;
        } catch (ReflectiveOperationException e) {
            throw new RuntimeException("Can't create a new instance of  " + clazz.getName());
        }
```

src/main/java/mate.academy/lib/Injector.java
```diff
@@ -8,6 +19,48 @@ public static Injector getInjector() {
     }
 
     public Object getInstance(Class<?> interfaceClazz) {
-        return null;
+        Class<?> clazz = findImplementation(interfaceClazz);
```
Elena-Bruyako, line 22: getImplementation
check Component presense
get fields
check Inject presense
create new instance
set

Elena-Bruyako (changes requested):
Good job! Let's improve your solution

PR #1273, opened 2026-08-29
initial commit

yenaofox:


yenaofox:
@mateacademy-ai-mentor
Injector.java file is located by path src/main/java/mate.academy/lib/Injector.java
where do you check this file?

